Issue #40, opened 2022-01-27
Typegen file not created on Windows


wKovacs64:
VSCode developer log for "Window" shows the following when saving a machine:

`[2022-01-27 16:34:50.561] [renderer1] [error] UNKNOWN: unknown error, stat '\\c%3A\dev\project\src\index.ts': Error: UNKNOWN: unknown error, stat '\\c%3A\dev\project\src\index.ts'`

And no typegen file is created. Probably a path resolution thing?

mattpocock:
OK, I know what's causing this, but need a bit of time to get set up on windows and fix it.

mattpocock:
For context, I'm pretty sure it's this line of code:

https://github.com/statelyai/xstate-vscode/blob/master/client/src/typegenService.ts#L57

batnyu:
Hello, 

I fixed the issue on Windows and made a pull request. #41 
Unfortunately, I don't have a Mac, so I was unable to test on macOs.
I hope it can help you!

Thanks!

mattpocock:
Fixed in 1.5.7

PR #41, opened 2022-01-28
fix: correct path on Windows to create and delete typegen file

batnyu:
Hello Matt,

Thanks for your work on XState, it's amazing and inspiring!

On Windows, the typegen file is not generated. I followed the lead you wrote in the issue #40.
I fixed the issue on Windows. Unfortunately, I don't have a Mac, so I was unable to test on macOs. But maybe, it can help you.

I used `vscode.Uri.parse` to extract the path from the URI. But the result is prefixed with a slash that caused `fs.writeFile` to prefix the path with "C:". So I removed this slash with `slice(1)` and it works! (at least on Windows).

Thanks!

### DIFF
--- a/client/src/typegenService.ts
+++ b/client/src/typegenService.ts
@@ -1,5 +1,5 @@
 import * as fs from "fs";
-import * as path from "path";
+import * as os from "os";
 import * as prettier from "prettier";
 import { promisify } from "util";
 import * as vscode from "vscode";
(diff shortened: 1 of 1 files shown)


client/src/typegenService.ts
```diff
@@ -49,14 +49,13 @@ const throttledTypegenCreationMachine = createMachine<
         await Promise.all([
           Object.entries(context.eventMap).map(async ([, event]) => {
             const uri = event.uri;
-
-            const newUri = vscode.Uri.file(
-              uri.replace(/\.([j,t])sx?$/, ".typegen.ts"),
+            const pathFromUri = vscode.Uri.parse(uri, true).path.slice(1);
```
Andarist, line 52: Could you tell us how does your `uri` exactly looks like? this could help us reviewing this change and comparing with unix-style paths

batnyu: You can see the different variables in this image : 

image

batnyu: When I remove the `slice(1)` from line 52, it throws an error when trying to write the file. `C:` is added somehow. 

image

batnyu: I tried my changes on a debian OS. It didn't work because I was removing the leading slash. I added a check to remove the leading slash only on Windows. I don't know if there is a cleaner way to handle this but now it works for both Windows and Unix

mattpocock:
@batnyu Tested this locally on Mac OS, and it works! Since this is a critical fix for Windows, let's merge it and figure out Unix later.

PR #51, opened 2022-02-09
Added XState CLI for typegen

mattpocock:


cli/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "xstate-cli",
+  "version": "0.0.1",
+  "bin": {
+    "xstate": "./bin/index.js"
```
Andarist, line 5: IMHO reusing "xstate" as a name for this can be slightly confusing, perhaps `xstate-cli` would avoid this problem (especially that it would match the package name)? 
```suggestion
    "xstate": "./bin/index.js"
```

cli/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "scripts": {
+    "prepare": "esbuild src/bin.ts --outfile=bin/index.js --bundle --platform=node --target=node14 --external:fsevents"
```
Andarist, line 8: is `fsevents` the only external that we want to keep as external?

cli/package.json
mattpocock: Or `stately`?

cli/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "xstate": "^4.29.0"
+  },
+  "peerDependencies": {
+    "xstate": "^4.29.0"
+  },
+  "dependencies": {
+    "chokidar": "^3.5.3",
+    "commander": "^8.0.0",
+    "fast-glob": "^3.2.11",
+    "xstate-parser-demo": "0.5.3",
```
Andarist, line 22: what's that? is this pkg really a dependency of the `xstate-cli`?

cli/package.json
mattpocock: Probably not, currently the bundle is like 17MB so something's going wrong. Probably an issue with preconstruct/dev and this?

cli/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "xstate-vscode-shared": "*"
+  },
+  "author": "Matt Pocock",
```
Andarist, line 25: perhaps smth like this would work better in the long run:
```suggestion
  "author": "Stately Team",
```

cli/package.json
mattpocock: Yes, we need to parse the code into machines. xstate-parser currently on npm as `xstate-parser-demo`, so that's why it's imported.

cli/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "peerDependencies": {
+    "xstate": "^4.29.0"
+  },
+  "dependencies": {
+    "chokidar": "^3.5.3",
+    "commander": "^8.0.0",
+    "fast-glob": "^3.2.11",
+    "xstate-parser-demo": "0.5.3",
+    "xstate-vscode-shared": "*"
```
Andarist, line 23: this perhaps should be restructured as it's rather weird that a cli depends on something that has vscode in the name

cli/package.json
Andarist: i might be wrong but it sounds like you are bundling the whole thing, while you might only want to bundle what's inside your source files, leaving all of the deps and peer deps as external

cli/package.json
Andarist: ah, OK - the package name threw me off

cli/package.json
Andarist: it depends what other responsibilities will be added here but for now the VScode extension is under th XState umbrella and not under Stately, so I would say that this CLI should match it for now

cli/src/bin.ts
```diff
@@ -0,0 +1,144 @@
...
+import { parseMachinesFromFile } from "xstate-parser-demo";
+import {
+  doesTsTypesRequireUpdate,
+  getRangeFromSourceLocation,
+  makeXStateUpdateEvent,
+  writeToTypegenFile,
+} from "xstate-vscode-shared";
+
+const readFile = promisify(fs.readFile);
+const writeFile = promisify(fs.writeFile);
```
farskid, line 18:
```suggestion
const readFile = fs.promises.readFile;
const writeFile = fs.promises.writeFile;
```

cli/src/bin.ts
```diff
@@ -0,0 +1,144 @@
...
+    const lengthDifference = fileEdit.newText.length - oldTextLength;
+    futureFileEdits.forEach((futureFileEdit) => {
+      if (futureFileEdit.start > fileEdit.end) {
+        futureFileEdit.start += lengthDifference;
+        futureFileEdit.end += lengthDifference;
+      }
+    });
+  }
+  return newText;
+};
```
farskid, line 50: Can we please unit test this bit?

cli/package.json
Andarist, line 8: Perhaps this would work:
```suggestion
    "prepare": "esbuild src/**/*.ts --outdir=bin --platform=node --target=node14"
```

cli/src/bin.ts
```diff
@@ -0,0 +1,144 @@
...
+          if (requiresUpdate) {
+            fileEdits.push({
+              start: machine.ast.definition.tsTypes.node.start!,
+              end: machine.ast.definition.tsTypes.node.end!,
+              newText: `{} as import("./${relativePath}.typegen").Typegen${machineIndex}`,
+            });
+          }
+          machineIndex++;
+        }
+      }
```
farskid, line 101: How is this editing different from the one above?

cli/src/bin.ts
```diff
@@ -0,0 +1,144 @@
...
+          parseResult: machine,
+        })),
+      );
+
+      const fileEdits: FileEdit[] = [];
+
+      for (const machine of parseResult.machines) {
+        let machineIndex = 0;
+        if (machine.ast.definition?.tsTypes?.node) {
+          const relativePath = removeExtension(path.basename(uri));
```
farskid, line 84: You can use `path.parse` built-in into Node.

cli/readme.md
```diff
@@ -0,0 +1,21 @@
+## Commands
+
+### xstate typegen <files>
+
+`xstate typegen "**/**/.ts"`
```
Andarist, line 5: This wouldn't find anything 😉 
```suggestion
`xstate typegen "src/**/*.ts"`
```

shared/src/writeToTypegenFile.ts
```diff
@@ -0,0 +1,34 @@
...
+}) => {
+  const prettierConfig = await prettier.resolveConfig(opts.filePath);
+  const pathToSave = opts.filePath.replace(
+    /\.(mjs|js|tsx|ts|jsx)$/,
+    ".typegen.ts",
+  );
+
+  try {
+    if (
+      opts.event.machines.filter((machine) => machine.hasTypesNode).length > 0
```
farskid, line 18:
```suggestion
      opts.event.machines.some((machine) => machine.hasTypesNode)
```

shared/src/writeToTypegenFile.ts
```diff
@@ -0,0 +1,34 @@
...
+    ) {
+      const typegenOutput = getTypegenOutput(opts.event);
+      await promisify(fs.writeFile)(
```
farskid, line 21:
```suggestion
      await fs.promises.writeFile(
```

farskid (changes requested):
Left some suggestions.

cli/src/bin.ts
Andarist: We are already targeting node 14 here so we can just do this:
```js
import fs from 'fs/promises'
```

farskid: Didn't suggest that, as I thought he might be using something in `fs` that isn't available in `fs/promises`

cli/package.json
mattpocock: Yess, that makes sense

cli/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "strictNullChecks": true,
+    "strict": true,
```
Andarist, line 4: `strictNullChecks` is already part of the `strict` "family"
```suggestion
    "strict": true,
```

cli/src/bin.ts
farskid: never mind. I thought we have 2 edits.

cli/src/bin.ts
```diff
@@ -0,0 +1,142 @@
...
+     * by the length of the new text MINUS the length of the old
+     */
+    const oldTextLength = fileEdit.end - fileEdit.start;
+    const lengthDifference = fileEdit.newText.length - oldTextLength;
+    futureFileEdits.forEach((futureFileEdit) => {
+      if (futureFileEdit.start > fileEdit.end) {
+        futureFileEdit.start += lengthDifference;
+        futureFileEdit.end += lengthDifference;
+      }
+    });
```
Andarist, line 46: Perhaps instead of this, you could sort file edits in the reversed order and just apply them from the last one to the first one. This potentially assumes that text edits can't be overlapping but I think this is sort of an invariant of this anyway - I don't see a situation when we'd like to create overlapping text edits

cli/package.json
mattpocock: Fixed, now down to 4.8kb.

cli/src/bin.ts
mattpocock: Urgh, that's so obvious and so much simpler. Great idea!

cli/package.json
mattpocock: Might be one for @davidkpiano, since it's technically branding

cli/src/bin.ts
```diff
@@ -0,0 +1,142 @@
...
+    await writeToFiles(allFiles);
+
+    if (opts.watch) {
+      watch(files, {}).on("change", (path) => {
+        if (path.endsWith(".typegen.ts")) {
+          return;
+        }
+        writeToFiles([path]);
+      });
+    }
```
Andarist, line 101: There are a couple of small problems with this:
1. the used glob matcher is different between the initial scan and the watcher
2. we might miss changes between the initial scan and setting up the watcher
3. we don't handle newly created files in the watcher
4. we don't clean up the outdated typegen files when a file is being removed (this probably requires more thought and shouldn't be part of this PR but it would be great to create an issue to handle this in the future)
5. there is a potential race condition here because we might receive `change` event before the file in question gets fully updated

How to fix those problems?
ad 1. reuse the pattern matching from chokidar - we can call `getWatched()` on the created watcher, it's not populated synchronously though so we should instantiate the watcher (even when we are not running in the watch mode), wait for the `ready` event and call it from there
ad 2. by using the first suggestion the problem should go away or should be minimized. Hard to tell which one without diving into chokidar's internals but my hope is that we could basically ignore all of the events arriving before the `ready` event. I think that the description of the `ignoreInitial` option implies that it's safe to do so. Alternatively, we should only listen for the `ready` event in the non-watch mode and process events asap when we receive them in the watch mode - this probably would be the best as we wouldn't even have to call `getWatched` in the watch mode. We'd just be notified about "adds" during the initial scan. When I think about it - we'd probably do the same in non-watch mode and just mark the single-pass scan as complete from within the `ready` event.
ad 3. kinda mentioned above - just listen for add events
ad 5. we could use `awaitWriteFinish: true`

There are also race conditions here for our own processing - I think that it would be best to ensure that they are impossible by queuing any new notifications related to the same file while our processing for a given file is already in progress.

cli/package.json
davidkpiano: We can move this to `@xstate/parser`

cli/src/bin.ts
mattpocock: @Andarist Any chance you could PR a change to this branch? I think I'd understand better by reading your code, and you appear to understand a lot more about chokidar than I do.

cli/package.json
davidkpiano: If it is under `xstate`, we should be maintaining it in the `statelyai/xstate` monorepo, under `packages/cli` 🤔

mattpocock: Let's discuss in Discord

cli/package.json
Andarist: it should be something like `@xstate/machines-extractor`, `@xstate/miner`, or something like this - as calling it a parser is a tiny stretch

cli/src/bin.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const handleError = (uri: string, e: any) => {
+  if (e?.code === "BABEL_PARSER_SYNTAX_ERROR") {
+    console.log(`${uri} - syntax error, skipping`);
+  }
+};
+
+const writeToFiles = async (uriArray: string[]) => {
+  /**
+   * TODO - implement batching to speed things up
```
Andarist, line 29: This is just about wrapping a single iteration of the loop in a promise and calling Promise.all on the created promises, right? seems like an easy change to be included as part of this PR

shared/src/writeToTypegenFile.ts
```diff
@@ -0,0 +1,32 @@
+import * as fs from "fs";
+import * as prettier from "prettier";
+import { promisify } from "util";
+import { getTypegenOutput } from ".";
+
+export const writeToTypegenFile = async (opts: {
+  filePath: string;
+  event: Parameters<typeof getTypegenOutput>[0];
```
Andarist, line 8: q: any particular reason why this is called an event here? i looks like a pretty imperative call 🤔

shared/src/writeToTypegenFile.ts
```diff
@@ -0,0 +1,32 @@
...
+}) => {
+  const prettierConfig = await prettier.resolveConfig(opts.filePath);
+  const pathToSave = opts.filePath.replace(
+    /\.(mjs|js|tsx|ts|jsx)$/,
```
Andarist, line 12:
```suggestion
    /\.(cjs|mjs|js|tsx|ts|jsx)$/,
```

cli/src/bin.ts
mattpocock: Iiish, we don't want to run 40+ checks at the same time. Although maybe that's better than not having it.

shared/src/writeToTypegenFile.ts
mattpocock: Yeah it's just called an event elsewhere in the code (`XStateUpdateEvent` from VSCode) so I kept it here.

shared/src/writeToTypegenFile.ts
Andarist: Although - I'm unsure how this should work for non-TS files, does it even work right now? Technically we could try to process JS files and insert a JSDoc-style type back to the file but I somewhat assume that this is not supported now (and not a high priority for now)

shared/src/writeToTypegenFile.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (opts.event.machines.some((machine) => machine.hasTypesNode)) {
+      const typegenOutput = getTypegenOutput(opts.event);
+      await promisify(fs.writeFile)(
+        pathToSave,
+        prettier.format(typegenOutput, {
+          ...prettierConfig,
+          parser: "typescript",
+        }),
+      );
+    } else if (await promisify(fs.exists)(pathToSave)) {
```
Andarist, line 26: Using `fs.exists` is deprecated and not recommended. The "blessed" way of doing most of the file-operations like this is to just do the thing and swallow the `ENOENT` error if it happens, while usually rethrowing other error types

shared/src/processFileEdits.ts
```diff
@@ -0,0 +1,23 @@
...
+  start: number;
+  end: number;
+  newText: string;
+};
+
+export const processFileEdits = (
+  oldText: string,
+  fileEdits: FileEdit[],
+): string => {
+  const fileEditsSortedByStart = fileEdits.sort((a, b) => b.start - a.start);
```
Andarist, line 11: I think this won't be a performance concern here and IMHO using an immutable update makes it usually easier to debug things as we can always inspect the previous state of a structure in the call stack etc
```suggestion
  const fileEditsSortedByStart = [...fileEdits].sort((a, b) => b.start - a.start);
```

shared/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "xstate-vscode-shared",
+  "name": "xstate-tools-shared",
```
Andarist, line 2:
```suggestion
  "name": "@xstate/tools-shared",
```

cli/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "peerDependencies": {
+    "xstate": "^4.29.0"
+  },
+  "dependencies": {
+    "chokidar": "^3.5.3",
+    "commander": "^8.0.0",
+    "fast-glob": "^3.2.11",
+    "xstate-parser-demo": "0.5.3",
+    "xstate-tools-shared": "*"
```
Andarist, line 23: Since we plan to release this we shouldn't use `*` here - it should use a more strict dependency range.

cli/src/bin.ts
Andarist: Happy to do that tomorrow 👍

cli/src/bin.ts
Andarist: Ye, not feeling super strong about it - this could always be chunked/backpressured if we see that too much parallelization is not that good. I'm happy with the current waterfall solution for now though. I've just thought that maybe you want to act on this comment soon(ish)

cli/package.json
mattpocock: Done in a future PR to this branch

shared/src/writeToTypegenFile.ts
```diff
@@ -0,0 +1,36 @@
...
+  const prettierConfig = await prettier.resolveConfig(opts.filePath);
+  const pathToSave = opts.filePath.replace(
+    /\.(cjs|mjs|js|tsx|ts|jsx)$/,
+    ".typegen.ts",
+  );
+
+  try {
+    if (opts.event.machines.some((machine) => machine.hasTypesNode)) {
+      const typegenOutput = getTypegenOutput(opts.event);
+      await promisify(fs.writeFile)(
```
farskid, line 19: This one's left behind.

farskid (approved):
The code looks good. Two minor comments only (nit-level).

PR #58, opened 2022-02-11
Tweak how we integrate with Chokidar

Andarist:
Partially addressing the points raised here: https://github.com/statelyai/xstate-vscode/pull/51/files#r802617971

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -2,10 +2,10 @@
   "name": "xstate-cli",
   "version": "0.0.1",
   "bin": {
-    "xstate": "./bin/index.js"
+    "xstate": "./bin/bin.js"
   },
   "scripts": {
-    "prepare": "esbuild src/bin.ts --outfile=bin/index.js --bundle --platform=node --external:./node_modules/* --external:../node_modules/*"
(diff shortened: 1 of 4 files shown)


cli/package.json
```diff
@@ -5,7 +5,7 @@
     "xstate": "./bin/index.js"
   },
   "scripts": {
-    "prepare": "esbuild src/bin.ts --outfile=bin/index.js --bundle --platform=node --external:./node_modules/* --external:../node_modules/*"
+    "prepare": "esbuild src/bin.ts --outdir=bin --platform=node --format=cjs"
```
Andarist, line 8: The previous script was not correct because it left relative paths to node_modules in the output

package.json
```diff
@@ -161,5 +161,6 @@
     "turbo": "^1.1.2",
     "typescript": "^4.3.2",
     "xstate-parser-demo": "0.5.3"
-  }
+  },
+  "prettier": {}
```
Andarist, line 165: This just helps my IDE to auto-format with Prettier because thanks to that it recognizes that this project is actually using Prettier.

mattpocock (approved):
Beautiful

PR #59, opened 2022-02-11
Inline implementations POC

mattpocock:


client/src/editorWebviewScript.ts
```diff
@@ -82,13 +88,21 @@ const machine = createMachine<WebViewMachineContext, EditorWebviewScriptEvent>(
       index: 0,
       layoutString: undefined,
       token: undefined,
+      implementations: {
+        implementations: {
```
davidkpiano, line 92: Same question as https://github.com/statelyai/registry/pull/505/files#r816433023 - why is `implementations` here twice?

mattpocock: Same answer, will refactor away

parser/src/machineCallExpression.ts
```diff
@@ -0,0 +1,46 @@
...
+  typeof MachineCallExpression
+>;
+
+export const MachineCallExpression = createParser({
+  babelMatcher: t.isCallExpression,
+  parseNode: (node, context) => {
+    if (
+      t.isMemberExpression(node.callee) &&
+      t.isIdentifier(node.callee.property) &&
+      ["createMachine", "Machine"].includes(node.callee.property.name)
```
davidkpiano, line 18: Thought: would it be simpler if we only supported `createMachine`? `Machine` is deprecated

parser/src/meta.ts
```diff
@@ -0,0 +1,9 @@
+import { StringLiteral, TemplateLiteral } from "./scalars";
+import { unionType } from "./unionType";
+import { objectTypeWithKnownKeys } from "./utils";
+
+export const MetaDescription = unionType([StringLiteral, TemplateLiteral]);
+
+export const StateMeta = objectTypeWithKnownKeys({
+  description: MetaDescription,
```
davidkpiano, line 8: Is this on `state.meta`? `description` is now a direct property of the state node, and we shouldn't assume it's on `meta.description` anymore... or maybe we still should. Thoughts welcome

davidkpiano (approved):
Code generally LGTM

client/src/editorWebviewScript.ts
```diff
@@ -82,6 +88,11 @@ const machine = createMachine<WebViewMachineContext, EditorWebviewScriptEvent>(
       index: 0,
       layoutString: undefined,
       token: undefined,
+      implementations: {
```
Andarist, line 91: q: any special reason why delays are not here?

parser/src/__tests__/choose.test.ts
```diff
@@ -0,0 +1,32 @@
+import { groupByUniqueName, parseMachinesFromFile } from "..";
+
+describe("Choose parsing", () => {
+  it("Should be able to grab actions and guardsdeclared inside a choose action", () => {
+    const result = parseMachinesFromFile(`
+      createMachine({
+        entry: [
+          choose([
```
Andarist, line 8: I wonder - one problematic pattern with choose is that `choose` could actually be declared within the implementations argument. Is such a situation handled?

parser/src/__tests__/invoke.test.ts
```diff
@@ -0,0 +1,41 @@
+import { parseMachinesFromFile } from "../parseMachinesFromFile";
+import { hashedId } from "../utils";
+
+describe("Invoke", () => {
+  it("Should allow for JSON stringifying anonymous invocations", () => {
+    const id = hashedId("() => {}");
+    const result = parseMachinesFromFile(`
+      createMachine({
+        invoke: {
+          src: "${id}",
```
Andarist, line 10: is this testing a correct thing though? this src here is not "anonymous" - it just has a very not friendly name 😉

parser/src/__tests__/parseResult.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    const machine = result.machines[0];
+
+    expect(machine.getAllStateNodes()).toHaveLength(4);
+
+    const aNode = machine.getStateNodeByPath(["a"]);
+
+    expect(aNode?.path).toEqual(["a"]);
+
+    const b1Node = machine.getStateNodeByPath(["b", "b1"]);
```
Andarist, line 27: q: what this is useful for?

parser/src/actions.ts
```diff
@@ -0,0 +1,310 @@
...
+    },
+  }),
+);
+
+export const ActionAsFunctionExpression = maybeTsAsExpression(
+  maybeIdentifierTo(
+    createParser({
+      babelMatcher: isFunctionOrArrowFunctionExpression,
+      parseNode: (node, context): ActionNode => {
+        const action = function actions() {};
```
Andarist, line 69: q: why do we need this? it's a dummy value so technically shouldn't be needed, right? couldn't the `action` property be optional? or part of a discriminated union like `NamedAction | InlineAction`?

parser/src/conds.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const CondAsFunctionExpression = createParser({
+  babelMatcher: isFunctionOrArrowFunctionExpression,
+  parseNode: (node, context): CondNode => {
+    return {
+      node,
+      name: "",
+      cond: () => {
+        return false;
+      },
```
Andarist, line 24: both of those feel a little bit redundant here

parser/src/scalars.ts
```diff
@@ -0,0 +1,89 @@
...
+export const TemplateLiteral = maybeTsAsExpression(
+  maybeIdentifierTo(
+    createParser({
+      babelMatcher: t.isTemplateLiteral,
+      parseNode: (node) => {
+        let value = "";
+
+        node.quasis.forEach((quasi) => {
+          value = `${value}${quasi.value.raw}`;
+        });
```
Andarist, line 81: this might lead to weird issues if there is actually more than a single quasi there - since those are interleaved by interpolations we compute a "weird" value here that doesn't really represent this template literal accurately.

parser/src/stateNode.ts
```diff
@@ -0,0 +1,114 @@
...
+  tags?: GetParserResult<typeof Tags>;
+  states?: GetParserResult<AnyParser<ObjectOfReturn<StateNodeReturn>>>;
+  node: t.Node;
+  meta?: GetParserResult<typeof StateMeta>;
+  tsTypes?: GetParserResult<typeof TsTypes>;
+  context?: GetParserResult<typeof Context>;
+  data?: GetParserResult<typeof AnyNode>;
+  schema?: GetParserResult<typeof Schema>;
+  preserveActionOrder?: GetParserResult<typeof BooleanLiteral>;
+  parallel?: GetParserResult<typeof BooleanLiteral>;
```
Andarist, line 65: This got superseded by `type: 'parallel'` quite a long time ago so we probably don't need to support this
```suggestion
```

parser/src/toMachineConfig.ts
```diff
@@ -0,0 +1,230 @@
...
+import { MaybeArrayOfActions } from "./actions";
+import { TMachineCallExpression } from "./machineCallExpression";
+import { StateNodeReturn } from "./stateNode";
+import { MaybeTransitionArray } from "./transitions";
+import { GetParserResult } from "./utils";
+
+export interface ToMachineConfigParseOptions {
+  /**
+   * Whether or not to hash inline implementations, which
+   * allow for parsing inline implementations as code.
```
Andarist, line 16: q: when do we not want this?

parser/src/toMachineConfig.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  return config;
+};
+
+export const toMachineConfig = (
+  result: TMachineCallExpression,
+  opts?: ToMachineConfigParseOptions,
+): MachineConfig<any, any, any> | undefined => {
+  if (!result?.definition) return undefined;
+  return parseStateNode(result?.definition, opts);
```
Andarist, line 167: q: what the intermediate representation is good for? why don't we create the config while visiting the AST?

shared/src/getInlineImplementations.ts
```diff
@@ -1,8 +1,10 @@
-import type { MachineParseResult } from "xstate-parser-demo/lib/MachineParseResult";
+import type { MachineParseResult } from "xstate-parser-demo/src/MachineParseResult";
+import { getRawTextFromNode, ImplementationsMetadata } from ".";
 
 export const getInlineImplementations = (
```
Andarist, line 4: could we document the idea behind this implementation?

shared/src/getNewMachineText.ts
```diff
@@ -0,0 +1,135 @@
...
+const prettierStartRegex = /^([^{]){1,}/;
+const prettierEndRegex = /([^}]){1,}$/;
+
+const UNWRAP_START = `@UNWRAP_START@`;
+const UNWRAP_END = `@UNWRAP_END@`;
+const markAsUnwrap = (str: string) => {
+  return `${UNWRAP_START}${str}${UNWRAP_END}`;
+};
+
+const UNWRAPPER_REGEX = /"@UNWRAP_START@(.{1,})@UNWRAP_END@"/g;
```
Andarist, line 16: this stuff really could use a comment

shared/src/getNewMachineText.ts
```diff
@@ -0,0 +1,135 @@
...
+      }
+
+      return value;
+    },
+    2,
+  );
+
+  const prettierConfig = await prettier.resolveConfig(fileName);
+
+  let finalTextToInput = `${json.slice(0, 1)}${nodesToPreserve.join(
```
Andarist, line 110: what this slicing is about? looks like perhaps it's grabbing `{` first, then inserting stuff that has been preserved, then inserting what got computed, and then ending with `}`. That would suggest that this can be changed like this to improve readability and symmetry with the closing curly brace:
```suggestion
  let finalTextToInput = `{${nodesToPreserve.join(
```

shared/src/getNewMachineText.ts
```diff
@@ -0,0 +1,135 @@
...
+        key === "cond" &&
+        implementations?.guards?.[value]?.jsImplementation
+      ) {
+        return markAsUnwrap(
+          implementations?.guards?.[value]?.jsImplementation!,
+        );
+      }
+      if (
+        key === "src" &&
+        implementations?.services?.[value]?.jsImplementation
```
Andarist, line 78: I wonder if instead of grabbing those texts first, then wrapping in special delimiters, then unwrapping etc we couldn't just pass the `oldText` here, keep ranges for those implementations and just slice this stuff from the original here.

shared/src/getNewMachineText.ts
```diff
@@ -0,0 +1,135 @@
...
+    // +1 and -1 for the quotes
+    return str
+      .slice(UNWRAP_START.length + 1, -UNWRAP_END.length - 1)
+      .replace(/\\n/g, "\n")
+      .replace(/\\"/g, '"')
+      .replace(/\\t/g, "\t");
+  });
+
+  try {
+    const result = await prettier.format(`(${finalTextToInput})`, {
```
Andarist, line 122: idea for VSCode: it's potentially possible to invoke "format selection"-like stuff after writing to the file, that way we could not depend on Prettier and just use what is configured by the IDE

parser/src/toMachineConfig.ts
mattpocock: For traversing the AST and holding meta information about how it's defined. It also exposes many helper methods for easier introspection/traversal. See `MachineParseResult.ts`

shared/src/getNewMachineText.ts
mattpocock: Makes sense!

shared/src/getNewMachineText.ts
mattpocock: Could you go into more detail?

shared/src/getNewMachineText.ts
mattpocock: Yes, I found this difficult to implement - it wasn't always working as expected.

parser/src/toMachineConfig.ts
mattpocock: We don't hash inline implementations for the viz, since we don't support them (and they're not necessary) so they'd show a bunch of horrible `123jha` ids in the actions/services.

parser/src/stateNode.ts
mattpocock: We HAVE to support everything that's in the TS declaration, otherwise TS yells at us.

parser/src/scalars.ts
mattpocock: True, though this hasn't caused an issue so far. Might be worth a TODO?

parser/src/actions.ts
mattpocock: It's a slight hangover from the way I've been testing this:

https://www.loom.com/share/bf8a1d808e8743d3b7f25ed97fa49c1f

parser/src/__tests__/invoke.test.ts
mattpocock: Yes, because the name MUST be deterministic. We're basically testing if hashedId ends up in the computed config.

parser/src/conds.ts
mattpocock: See comment above on the weird way I've been testing this

parser/src/machineCallExpression.ts
mattpocock: No, it wouldn't be simpler, especially since the code is already written. We should be able to parse anything the user gives us.

parser/src/meta.ts
mattpocock: We need to be able to parse whatever the user throws at us - the parser should handle not just current XState syntax, but also slightly older stuff too.

parser/src/scalars.ts
```diff
@@ -0,0 +1,89 @@
...
+});
+
+export const TemplateLiteral = maybeTsAsExpression(
+  maybeIdentifierTo(
+    createParser({
+      babelMatcher: t.isTemplateLiteral,
+      parseNode: (node) => {
+        let value = "";
+
+        node.quasis.forEach((quasi) => {
```
mattpocock, line 79:
```suggestion
        // TODO - this might lead to weird issues if there is actually more than a single quasi there
        node.quasis.forEach((quasi) => {
```

parser/src/__tests__/choose.test.ts
mattpocock: No - it isn't, I'll add an issue:

https://github.com/statelyai/xstate-vscode/issues/70

client/src/editorWebviewScript.ts
mattpocock: I hadn't thought that inline delays were possible. Farzad mentioned that they are, but I don't see much usage for them. We can add them later:

https://github.com/statelyai/xstate-vscode/issues/71

Andarist: is there any significant cost to supporting them right away though?

parser/src/__tests__/parseResult.test.ts
Andarist: Interesting. Could you describe this in more detail? Like what's the exact user story for this?

parser/src/scalars.ts
mattpocock: Done

parser/src/toMachineConfig.ts
Andarist: What does it mean for the viz though? Can machines with inline implementation be visualized? Couldn't we just display those as "anonymous" in the viz?

parser/src/__tests__/parseResult.test.ts
mattpocock: The way we store the AST uses a recursive structure, similar to XState's config itself:

```
// pseudo-code
definition: {
  states: {
    a: {
      states: {}
    }
  }
}
```

This is a useful helper for when you need to traverse deep in the state. It also means we can directly interop between XState's paths and the AST.

The user story is:

WHEN a user CMD-clicks a target
THEN they should be able to jump to the targeted node via jump-to-definition

client/src/editorWebviewScript.ts
mattpocock: Just the fact that I'd need to run the changes through this chain of 4 PR's, which I don't want to do.

mattpocock:
FYI, building this branch into a VSIX is broken, but it's fixed on a future PR

shared/src/getNewMachineText.ts
Andarist: Instead of using `JSON.stringify` we could just traverse the `newConfig` (after sorting its keys) on our own and stringify stuff as we go. If we'd encounter an inline action/whatever then we could look it by the hashed id in the bookkept stuff where the shape of bookkept information would look something like:
```ts
type InlineImplementation = {
  type: 'inline',
  range: {
    start: number;
    end: number;
  }
}
```

Based on this we could just stringify it directly as:
```js
const implementation = implementations[hashedId]
const str = oldText.slice(implementation.range.start, implementation.range.end)
```

parser/src/toMachineConfig.ts
mattpocock: I remember why I didn't do this here - again, it's to do with the existing test setup in @xstate/machine-extractor. Let's address this as a followup PR?

shared/src/getNewMachineText.ts
mattpocock: Let's add this as a followup PR.

mattpocock: Summary from call: we're dissatisfied in general with JSON.stringify and want to find a better solution, but not now.

Andarist (approved):

PR #68, opened 2022-02-28
Refactored to monorepo

mattpocock:
This PR does a load of things:

1. Refactored vscode build process to esbuild, removed an old dependency on `ncc`.
2. Renames `xstate-parser-demo` to `@xstate/machine-extractor`
3. Adds changesets, including the changesets GitHub action.
4. Gets all the scripts ready for deploying to `@xstate/*`


.changeset/config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@1.7.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "fixed": [],
+  "linked": [],
+  "access": "restricted",
+  "baseBranch": "master",
```
Andarist, line 8: worth renaming the base branch to `main` in this repo

.github/workflows/ci.yml
```diff
@@ -16,18 +16,8 @@ jobs:
 
       - run: yarn
       - run: yarn turbo run lint test
-      - run: npx vsce package
+      - run: yarn deps:build && yarn extension:compile && (cd apps/extension/client && npx vsce package --yarn)
```
Andarist, line 19: nit: splitting into separate runs would end up with a somewhat nicer log and more focused reports when some part of this fails

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+      - "master"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
Andarist, line 13: this is an outdated recommendation from the Changesets action - we now automatically "deepen" the git tree so it shouldn't be required to fetch all the git history by default
```suggestion
```

I probably should remove that from the `changesets/action`'s README :sweat_smile:

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+      - run: yarn
+      - run: yarn deps:build && yarn extension:compile
+      - name: Create Release Pull Request or Publish to npm
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          publish: yarn release
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          VSCE_PAT: ${{ secrets.VSCE_PAT }}
```
Andarist, line 28: We probably also need the `GITHUB_TOKEN` here to create github releases etc

.changeset/config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@1.7.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
```
Andarist, line 3: I recommend using the GitHub generator, like here:
https://github.com/statelyai/xstate/blob/bc592582a8b486ca3a6277ca9a9c922590572346/.changeset/config.json#L3

This package would have to be installed for this work.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14.x
+          cache: "yarn"
+
+      - run: yarn
+      - run: yarn deps:build && yarn extension:compile
+      - name: Create Release Pull Request or Publish to npm
```
Andarist, line 21:
```suggestion
      - name: Create Release Pull Request or Publish
```

apps/cli/package.json
```diff
@@ -1,11 +1,12 @@
 {
-  "name": "xstate-cli",
+  "name": "@xstate/cli",
   "version": "0.0.1",
   "bin": {
     "xstate": "./bin/bin.js"
```
Andarist, line 5: WDYT about making this `xstate-cli` to differentiate from the core?

.github/workflows/publish.yml
mattpocock: Do I need to manually add the GITHUB_TOKEN to the secrets, or is it available?

apps/extension/client/package.json
```diff
@@ -0,0 +1,96 @@
+{
+  "name": "stately-vscode",
+  "displayName": "XState VSCode",
+  "description": "Visual editing, autocomplete and typegen for XState",
+  "license": "MIT",
+  "author": "Stately Team",
+  "version": "1.6.1",
```
Andarist, line 7: I'm not completely sure how this will behave when publishing to the extension store etc but from the PoV of Changesets+npm it's probable safer to start with this:

```suggestion
  "version": "1.6.1",
  "private": true
```

.github/workflows/publish.yml
Andarist: It's available by default, but you need to share it explicitly with the action by adding it to `env` here

packages/shared/src/types.ts
```diff
@@ -1,5 +1,5 @@
 import { MachineConfig, StateMachine } from "xstate";
-import { MachineParseResult } from "xstate-parser-demo/src/MachineParseResult";
+import { MachineParseResult } from "@xstate/machine-extractor/src/MachineParseResult";
```
Andarist, line 2: shouldn't this be imported from a built dist file? so probably like this:
```suggestion
import { MachineParseResult } from "@xstate/machine-extractor";
```

packages/shared/src/getNewMachineText.ts
```diff
@@ -1,6 +1,6 @@
 import * as prettier from "prettier";
 import { MachineConfig } from "xstate";
-import type { MachineParseResult } from "xstate-parser-demo/src/MachineParseResult";
+import type { MachineParseResult } from "@xstate/machine-extractor/src/MachineParseResult";
```
Andarist, line 3: Potentially:
```suggestion
import type { MachineParseResult } from "@xstate/machine-extractor";
```

packages/shared/src/getInlineImplementations.ts
```diff
@@ -1,4 +1,4 @@
-import type { MachineParseResult } from "xstate-parser-demo/src/MachineParseResult";
+import type { MachineParseResult } from "@xstate/machine-extractor/src/MachineParseResult";
```
Andarist, line 1: Potentially:
```suggestion
import type { MachineParseResult } from "@xstate/machine-extractor";
```

packages/shared/src/__tests__/getNewMachineText.test.ts
```diff
@@ -1,7 +1,7 @@
-import { parseMachinesFromFile } from "xstate-parser-demo";
+import { parseMachinesFromFile } from "@xstate/machine-extractor";
 import { getNewMachineText, ImplementationsMetadata } from "..";
 import { MachineConfig } from "xstate";
-import { hashedId } from "xstate-parser-demo/src/utils";
+import { hashedId } from "@xstate/machine-extractor/src/utils";
```
Andarist, line 4: this potentially should also be imported from the root, or perhaps this util should be moved to shared utils?

packages/machine-extractor/package.json
```diff
@@ -1,14 +1,13 @@
...
-  "main": "dist/xstate-parser-demo.cjs.js",
+  "main": "dist/xstate-machine-extractor.cjs.js",
   "author": "Matt Pocock",
   "license": "MIT",
   "scripts": {
     "test": "jest src",
     "lint": "tsc",
-    "prepare": "patch-package && yarn build",
-    "build:prod": "(cd ../ && preconstruct build)"
+    "prepare": "patch-package && yarn build"
```
Andarist, line 10: this patch is most likely not relevant for the nested package - usually we only want such things at the root, so perhaps this is outdated and not needed at all?

Andarist: `build` script is also not defined here so probably this `prepare` script can be dropped

packages/machine-extractor/example-site/package.json
```diff
@@ -3,14 +3,14 @@
   "version": "0.0.0",
   "scripts": {
     "dev": "next",
-    "build": "yarn add xstate-parser-demo@latest && yarn && next build",
+    "build": "yarn add @xstate/machine-extractor@latest && yarn && next build",
```
Andarist, line 6: q: why do you install a package here? 🤔

packages/machine-extractor/example-site/package.json
```diff
@@ -3,14 +3,14 @@
...
     "serve": "next serve"
   },
   "dependencies": {
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
     "xstate": "^4.24.1",
-    "xstate-parser-demo": "^0.1.4",
+    "@xstate/machine-extractor": "^0.1.4",
```
Andarist, line 13: looks like this `package.json` might get updated in some future PR accidentally~ since those dependencies are not sorted

package.json
```diff
@@ -14,95 +11,34 @@
...
   "scripts": {
-    "cli": "yarn workspace xstate-cli",
-    "parser": "yarn workspace xstate-parser-demo",
-    "server": "yarn workspace xstate-vscode-server",
-    "client": "yarn workspace xstate-vscode-client",
-    "shared": "yarn workspace xstate-tools-shared",
-    "vscode:prepublish": "yarn turbo run build:prod",
-    "compile": "yarn turbo run build:dev --no-cache",
+    "cli": "yarn workspace @xstate/cli",
+    "parser": "yarn workspace @xstate/machine-extractor",
```
Andarist, line 29:
```suggestion
    "extractor": "yarn workspace @xstate/machine-extractor",
```

apps/extension/server/src/server.ts
```diff
@@ -18,7 +18,7 @@ import {
   TextDocumentSyncKind,
 } from "vscode-languageserver/node";
 import { assign, createMachine, interpret } from "xstate";
-import { MachineParseResult } from "xstate-parser-demo/src/MachineParseResult";
+import { MachineParseResult } from "@xstate/machine-extractor/src/MachineParseResult";
```
Andarist, line 21: potentially
```suggestion
import { MachineParseResult } from "@xstate/machine-extractor";
```

apps/extension/server/src/getCursorHoverType.ts
```diff
@@ -1,12 +1,12 @@
 import * as t from "@babel/types";
 import { Position } from "vscode-languageserver-textdocument";
-import { StringLiteralNode } from "xstate-parser-demo";
-import { MachineParseResult } from "xstate-parser-demo/src/MachineParseResult";
-import { StateNodeReturn } from "xstate-parser-demo/src/stateNode";
+import { StringLiteralNode } from "@xstate/machine-extractor";
+import { MachineParseResult } from "@xstate/machine-extractor/src/MachineParseResult";
+import { StateNodeReturn } from "@xstate/machine-extractor/src/stateNode";
```
Andarist, line 5: potentially:
```suggestion
import { MachineParseResult, StateNodeReturn, StringLiteralNode } from "@xstate/machine-extractor";
```

apps/extension/server/package.json
```diff
@@ -12,15 +11,14 @@
...
-    "xstate": "^4.29.0",
-    "xstate-tools-shared": "*"
+    "xstate": "^4.29.0"
   },
   "devDependencies": {},
   "scripts": {
     "lint": "tsc",
-    "build:dev": "esbuild --bundle src/server.ts --outfile=dist/index.js --external:vscode --platform=\"node\"",
-    "build:prod": "ncc build src/server.ts -t -o dist"
+    "build": "esbuild --bundle src/server.ts --outfile=dist/index.js --external:vscode --external:prettier --platform=\"node\" --format=cjs"
```
Andarist, line 22: q: isn't this whole package part of the extension "package"? do we need multiple package.json files to describe a single extension? is this correctly setup right now with `--bundle` etc?

apps/cli/package.json
Andarist: Oh, I'm on board with `@xstate/cli` for the package name. I was asking about the name of this executable:
```suggestion
    "xstate-cli": "./bin/bin.js"
```

mattpocock: Ah, gotcha. @davidkpiano?

I like `xstate` - short and snappy.

packages/machine-extractor/example-site/package.json
mattpocock: Great question, this is from the example-site which can probably be dropped.

apps/cli/package.json
Andarist: Just two examples from the wild:

webpack is using webpack-cli
vue is using vue

davidkpiano: @mattpocock `xstate` is fine 👍

Andarist (approved):

PR #69, opened 2022-03-01
[Do not merge until parents get merged] Fixed issue where tags get erased

mattpocock:


davidkpiano (approved):

packages/machine-extractor/examples/tags.machine.ts
```diff
@@ -0,0 +1,13 @@
+import { createMachine } from "xstate";
+
+export const tagMachine = createMachine({
```
Andarist, line 3: are those files here like run over with the extractor and then written back and compared? in other words - is this acting as a test?

Andarist (approved):

packages/machine-extractor/examples/tags.machine.ts
mattpocock: Yes, exactly

PR #74, opened 2022-03-03
Access Babel stuff through `@babel/core`

Andarist:
Please review this with care, I've executed build, lint and test scripts and I think that it works OK - but I don't have enough knowledge about this repo to be 100% sure about this


mattpocock (approved):
Tested locally and the tests/linting passes, so looks great!

PR #75, opened 2022-03-03
Add a script for git tagging the vscode extension

Andarist:
Hopefully, I didn't do any silly mistake here - I've tried to test this is much as possible but the production usage will verify if this is correct ;p

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "shared": "yarn workspace @xstate/tools-shared",
     "prepare": "husky install",
     "postinstall": "preconstruct dev && manypkg check",
-    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && npx vsce publish --packagePath $(find . -iname *.vsix))",
+    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && npx vsce publish --packagePath $(find . -iname *.vsix)) && node ./scripts/tag-extension.js",
     "deps:build": "preconstruct build",
     "extension:compile": "yarn turbo run build --no-cache && cp apps/extension/server/dist/index.js apps/extension/client/dist/server.js"
   },
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "shared": "yarn workspace @xstate/tools-shared",
     "prepare": "husky install",
     "postinstall": "preconstruct dev && manypkg check",
-    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && npx vsce publish --packagePath $(find . -iname *.vsix)) && node ./scripts/tag-extension.js",
+    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && (npx vsce publish --packagePath $(find . -iname *.vsix)) || true) && node ./scripts/tag-extension.js",
     "deps:build": "preconstruct build",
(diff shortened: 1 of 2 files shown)


scripts/tag-extension.js
```diff
@@ -0,0 +1,30 @@
...
+  "extension",
+  "client",
+  "package.json"
+));
+
+const tag = `${pkgJson.name}@${pkgJson.version}`;
+
+const { status, stdout } = spawnSync("git", [
+  "ls-remote",
+  "https://github.com/statelyai/xstate-tools.git",
```
Andarist, line 18: note that this relies on this repo being renamed as planned

mattpocock: Can we keep it the current name now just for testing? That way I can rename it all in one go

package.json
```diff
@@ -32,7 +32,7 @@
     "shared": "yarn workspace @xstate/tools-shared",
     "prepare": "husky install",
     "postinstall": "preconstruct dev && manypkg check",
-    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && npx vsce publish --packagePath $(find . -iname *.vsix))",
+    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && npx vsce publish --packagePath $(find . -iname *.vsix)) && node ./scripts/tag-extension.js",
```
mattpocock, line 35: Should this be run after the changeset publish or before? I'm sure you thought of this.

scripts/tag-extension.js
Andarist, line 18:
```suggestion
  "https://github.com/statelyai/xstate-vscode.git",
```

Andarist: done

mattpocock (approved):

package.json
Andarist: it probably doesn't matter - if anything fails during this pipeline then the whole release script should fail and `changesets/action` shouldn't create any github releases (some packages could have been released during the process though, we don't receive "partial" successes anyhow to handle this any better).

The current order matches the behavior of Changesets CLI - it first tries to publish packages to npm and only when it's done with that step then it creates tags for the published packages.

mattpocock: OO, one thing we've missed here is swallowing the error from the vsce publish.

```suggestion
    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && npx vsce publish --packagePath $(find . -iname *.vsix)) || true && node ./scripts/tag-extension.js",
```

Andarist: when it can fail? the scenario I imagine in which we should swallow the error is when this particular version has already been published and their store reject us based on that - we shouldn't swallow in any other cases, right?

also - are you sure that this proposed `||` is correct? this probably needs some parens wrapped around this and vsce

mattpocock:
```suggestion
    "release": "(cd apps/extension/client && npx vsce package --yarn) && changeset publish && (cd apps/extension/client && (npx vsce publish --packagePath $(find . -iname *.vsix)) || true) && node ./scripts/tag-extension.js",
```

Andarist: > Yes, exactly. But how do we selectively swallow that error in bash?

No idea XD I try to stay away from Bash as much as I can :P I guess it's acceptable for now to just roll with this and we see if it ever fails for other reasons

mattpocock: Yeah it's much more likely to fail on build, which is a separate proc

PR #84, opened 2022-03-03
Fixed a bug where transition targets would not be parsed correctly if declared with a template literal

### LINKED ISSUE #63: Bug with visualiser when using literal template string
Hopefully all will be explained in the below screenshot:

image

Using the following code:

```ts
const machine = createMachine({
	id: 'foo',
	states: {
		a: {
			on: {
				a: `b`,
			},
		},
		b: {},
	},
});
```

Produces a state machine whose "b" event transition isn't connected in the visualiser.

mattpocock:
Fixes #63

### DIFF
--- a/.changeset/fluffy-moose-sin.md
+++ b/.changeset/fluffy-moose-sin.md
@@ -4,4 +4,6 @@
 "stately-vscode": patch
 ---
 
+author: @Andarist 
+
 Fixed a bug where transition targets would not be parsed correctly if they were declared using a template literal.


.changeset/fluffy-moose-sin.md
```diff
@@ -0,0 +1,7 @@
+---
+"@xstate/machine-extractor": patch
+"@xstate/cli": patch
+"stately-vscode": patch
+---
+
+Fixed a bug where transition targets would not be parsed correctly if they were declared using a template literal.
```
mattpocock, line 7: @Andarist this PR is based on your work. How do I mark you as the author in the changeset?

Andarist:
```suggestion
author: @Andarist 

Fixed a bug where transition targets would not be parsed correctly if they were declared using a template literal.
```

PR #86, opened 2022-03-03
Update `caniuse-lite` to avoid spammy logs about outdated data

Andarist:
The problem is that this log is displayed on the time diff between some release dates and `Date.now()`. So it can start being displayed any day again - I'm still looking if there is any way to disable this check altogether


mattpocock (approved):

Andarist:
The only way to disable this altogether is to use env var:
https://github.com/browserslist/browserslist/blob/3cba97f1eacb468b92b33acc6819624e5edd42b0/node.js#L375

but I think that this shouldn't impact our customers cause this should only be verified when using `targets` and we only use it here:
https://github.com/statelyai/xstate-tools/blob/791c2a7b596d9356f15f16e2c839f27bf918c566/packages/machine-extractor/babel.config.js#L3

So probably not worth disabling it and we can just update this database from time to time

Issue #92, opened 2022-03-04
CLI: Unable to specify multiple globs, and excessive logging


simonflk:
Hi,

Great work on the CLI, I'm really excited to use this, and this will help me finally get my typegen PR merged.

I just tried it out and ran into a couple of issues:

---
1. This is minor, but I could not get the example in the docs to work:

```bash
xstate typegen "src/**/*.tsx?"
```

I believe it needs to change to:

```bash
xstate typegen "src/**/*.ts?(x)"
```
---

2. Excessive logging

It looks out `${file} - success` for every matching file - even if the file does not contain any machines.

--- 

3.  I cannot specify multiple globs

I have a monorepo situation, and I want to be able to run typegen across the "apps" folder, and also the "libs" folder. I was trying this:

```bash
xstate typegen "apps/**/*.ts" "libs/**/*.ts"
```

However, only the 'apps' folder is touched.

I also tried brace expansion:

```bash
xstate typegen "{apps,libs}/**/*.ts"
```

but this is not supported by [picomatch](https://github.com/micromatch/picomatch#braces) (used by [chokidar](npm.im/chokidar))

I also tried

```bash
xstate typegen "**/*.ts"
```

however, this traverses everything, including node_modules.

Andarist:
Do you have an example of some other CLI that accepts multiple globs? I rarely need such a thing and would like to take a look at the prior art.

simonflk:
> Do you have an example of some other CLI that accepts multiple globs? I rarely need such a thing and would like to take a look at the prior art.

Good question. In most cases CLIs I use just accept a list of filenames and you can use multiple globs which will be resolved by your shell . E.g. `git add *json *ts`

But I think an example of a CLI that accepts and interprets multiple globs itself would be [prettier](https://prettier.io/docs/en/cli.html)

PR #93, opened 2022-03-04
Fixed example glob in the CLI's README

Andarist:
Related to https://github.com/statelyai/xstate-tools/issues/92


mattpocock (approved):
We'll also need to fix the one in the xstate docs, too

Andarist:
@mattpocock done

Issue #96, opened 2022-03-06
VSCode extentsion (1.7.2) does not show prompt for editor / inspector


mdpratt:
The ability to open the visual editor or inspector is no longer working. I've validated that the issue is present all 1.7.x installs. I've downgraded to 1.6.0 for the time being, which is still working.

**Version 1.6.0**
![image](https://user-images.githubusercontent.com/982135/156915861-6d6ff3a3-2cac-453f-98e2-a05ae822fb5d.png)

**Version 1.7.2**
![image](https://user-images.githubusercontent.com/982135/156915907-be488bf6-1f5b-47d6-b10b-3aa1f64681d5.png)

mattpocock:
@mdpratt Can't repeat this on my machine, could you do two things:

1. Give me info about your OS/node version
2. Check the 'output' for XState in VSCode here. You might be able to spot some errors.

![image](https://user-images.githubusercontent.com/28293365/157003424-5928300a-d541-4f69-9d98-6beef92a2748.png)

mdpratt:
I'm using Remote-SSH to connect to an AWS EC2 instance that I develop from. As such, I'm not seeing an XState output in my list. Looking through other logs, I'm seeing a few things:

**Log (Remote Exentsion)**
> [2022-03-07 11:26:52.546] [exthost] [info] ExtensionService#_doActivateExtension statelyai.stately-vscode, startup: false, activationEvent: 'onLanguage:typescript'

**Remote host:**
OS:
```bash
NAME="Amazon Linux"
VERSION="2"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2:-:"
HOME_URL="https://amazonlinux.com/"
```

* Node: v17.6.0
* NPM: 8.5.1

**Local host:**
* Windows 10 (Version 20H2, Build 19042.1526)
* Node: v17.6.0
* NPM: 8.5.1

mdpratt:
Downgrading to `v1.7.0` does produce this error though:

```bash
[2022-03-07 11:37:22.681] [exthost] [error] Activating extension statelyai.stately-vscode failed due to an error:
[2022-03-07 11:37:22.681] [exthost] [error] Error: Connection to server got closed. Server will not be restarted.
[...]

PR #97, opened 2022-03-09
Revert @babel/core changes

mattpocock:


Andarist:
@mattpocock could you comment on why this is needed and what kind of problems did this create? It's far from obvious now ;p

mattpocock:
@Andarist Looks like this makes the VSCode parsing fail on some machines:

#96

Andarist:
Do we have any more info about this or a repro case? The issue only mentions problems with finding the prettier module.

Andarist:
Could we perhaps gather more info? maybe release additional logs etc?  If we add those now they might both help us fix this and also help us diagnose future issues.

mattpocock:
@Andarist I think probably we should revert this change first (since it's an urgent fix for prod and we know it works) then investigate.

Andarist:
If we revert without investigating first... how do we investigate this later?

Andarist:
Either way - It's your call and I will totally respect your decision here. But there is a missing test case here (although it's dependent on the Babel implementation) that we are missing and we might weirdly end up with the same problem somehow in the future

mattpocock:
@Andarist Yep, I can see that. I think I'd prefer to know it's fixed than wonder why it's broken.

mattpocock:
@Andarist Actually, ignore above, I've managed to reproduce using MDPratt's machine

PR #100, opened 2022-03-10
Fixed a bug where inline guards were not being picked up in non-root events

mattpocock:


packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -22,7 +22,7 @@ export interface ToMachineConfigParseOptions {
 
 const parseStateNode = (
   astResult: StateNodeReturn,
-  opts?: ToMachineConfigParseOptions,
+  opts: ToMachineConfigParseOptions | undefined,
```
mattpocock, line 25: The issue was I wasn't passing the options to every function that needed it

packages/machine-extractor/src/__tests__/inline-implementations.test.ts
```diff
@@ -0,0 +1,28 @@
+import { parseMachinesFromFile } from "../parseMachinesFromFile";
+
+describe("Inline implementations", () => {
+  it("Should pick up guards expressed inline", () => {
```
mattpocock, line 4:
```suggestion
  it("Should pick up guards declared inline deeper than the root level", () => {
```

Andarist (approved):

packages/machine-extractor/src/toMachineConfig.ts
Andarist: That's a good trick to ensure that you always pass params around. I'm using this in Changesets in a couple of places too. 10/10, would recommend to a friend

PR #109, opened 2022-03-17
Fixed support for experimental `.cts` and `.mts` extensions

Andarist:


packages/shared/src/writeToTypegenFile.ts
```diff
@@ -7,10 +8,7 @@ export const writeToTypegenFile = async (opts: {
   event: Parameters<typeof getTypegenOutput>[0];
 }) => {
   const prettierConfig = await prettier.resolveConfig(opts.filePath);
-  const pathToSave = opts.filePath.replace(
-    /\.(cjs|mjs|js|tsx|ts|jsx)$/,
-    ".typegen.ts",
-  );
+  const pathToSave = opts.filePath.slice(0, -(path.extname(opts.filePath).length)) + ".typegen.ts"
```
Andarist, line 11: I've chosen this approach since we don't actually care about the `opts.filePath` extension. We just want to replace it. This also ensures that the `pathToSave` is always different than `opts.filePath`.

This should be compatible with `removeExtension` from [here](https://github.com/statelyai/xstate-tools/blob/9215fff37efdbacf6827cf38a4b7f0f3055b921c/apps/extension/client/src/initiateTypegen.ts#L93-L95). Note that we can't reuse this `removeExtension` here because it assumes that it operates on the basename and in here we have a full path~. Checking the last dot in the full path would come with an edge case bug as a dot could be in one of the path segments.

mattpocock (approved):
Nice, looks good!

PR #111, opened 2022-03-18
Made the base url configurable by the user

mattpocock:
This allows for changing the base url to localhost:3000 during development.

No changeset needed, since this change is most useful for local dev only.

### DIFF
--- a/apps/extension/client/package.json
+++ b/apps/extension/client/package.json
@@ -71,6 +71,11 @@
           "type": "boolean",
           "default": true,
           "description": "Show warnings that prevent you from using the visual editor."
+        },
+        "xstate.targetEditorBaseUrl": {
+          "type": "string",
+          "default": "https://stately.ai",
+          "description": "The base URL to open the visual editor in."
         }
       }
     }
--- a/apps/extension/client/package.json
+++ b/apps/extension/client/package.json
@@ -82,7 +82,7 @@
   },
   "scripts": {
     "lint": "tsc",
-    "test": "jest",
+    "test": "jest --passWithNoTests",
     "viz:build": "esbuild src/vizWebviewScript.ts --bundle --outfile=scripts/vizWebview.js",
     "editor:build": "esbuild src/editorWebviewScript.ts --bundle --outfile=scripts/editorWebview.js",
     "build": "esbuild --bundle --platform=\"node\" src/extension.ts --outfile=dist/index.js --external:vscode --format=cjs"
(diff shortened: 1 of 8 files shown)


apps/extension/client/package.json
```diff
@@ -71,6 +71,11 @@
           "type": "boolean",
           "default": true,
           "description": "Show warnings that prevent you from using the visual editor."
+        },
+        "xstate.targetEditorBaseUrl": {
+          "type": "string",
+          "default": "https://stately.ai",
+          "description": "The base URL to open the visual editor in."
```
Andarist, line 78: This makes it a public option and I think that it should be avoided. Couldn't you source this from a local configuration file or an environment variable to make this at least more hidden from the users?

mattpocock: Why do you think this should be avoided? I think it's fine to have it public. It makes it ultra-convenient in development.

Andarist: How an external user can really use this? How useful it is for them to see this in autocomplete when configuring the extension? It's just that IMHO if it's not useful for them then it shouldn't be there -> fewer options === less mental overhead.

Would a different place for configuring this locally be cumbersome to use?

davidkpiano: This is a typical place to put the setting IMO - other extensions do similar (these are from mine):

CleanShot 2022-03-18 at 07 47 17
.

Andarist: Right, this is a typical place to put a setting - I don't argue that. I just say that this is not a user-facing setting and as such it probably shouldn't be in such generated forms etc.

davidkpiano: In the future, we could use this to cater to users who want to self-host the editor/registry, or have it on some stately subdomain (`beta.stately.ai`) or other user-facing ideas.

Andarist: Isnt this potentially YAGNI right now? It's not hard to introduce this later after all, when needed. 

To clarify - this is a nit-level and of course just my opinion, feel free to ignore. Just from my PoV this kinda becomes the API surface when we make this a public option but I dont see how a regular user would use this *today* and therefore my preference is to hide this from this public API surface

mattpocock: It is potentially YAGNI, but it's very very useful in development and has little cost to add, so I figured I'd add it.

Andarist: But I'm not saying that we shouldn't have this configurable - just that perhaps a different way of configuring this during dev would be almost as good as this one and that an alternative way wouldn't expose this to our users, where I consider the latter to be a "hidden cost" of adding this as a public option.

Andarist (approved):

apps/extension/client/package.json
mattpocock: It's hard to imagine a better option than this during development - you can even change it while the server is running, or specify it differently in different repos with workspace-based config.

PR #112, opened 2022-03-18
Added a test suite for the CLI and fixed a bug where multiple machines would not work properly

mattpocock:


apps/cli/package.json
```diff
@@ -6,11 +6,15 @@
   },
   "scripts": {
     "prepare": "yarn build",
+    "lint": "tsc",
+    "test": "jest src",
```
Andarist, line 10: Is this position argument needed here? 🤔 
```suggestion
    "test": "jest",
```

apps/cli/src/program.ts
```diff
@@ -0,0 +1,133 @@
...
+        console.log(`${uri} - success`);
+      } catch (e) {
+        handleError(uri, e);
+      }
+    }),
+  );
+};
+
+export const runTypegen = (filesPattern: string, opts: { watch?: boolean }) => {
+  return new Promise<void>((resolve, reject) => {
```
Andarist, line 93: Returning a promise from something that is passed to the `.action` might have a special meaning there. This is a little bit iffy for me, especially given that only one specific branch here returns a promise. 

This could be restructured in a way that you'd export a function returning a promise, but one that wouldn't even deal with `chokidar`. So basically the inner `writeToFiles` could be renamed, promisified (if it's not already), and exported from here.

Andarist:
Since the file has been renamed there is no good diff there and it's not quite obvious what actually has been changed in the implementation and where and how the bug was fixed. Could you add a PR comment near the related lines?

apps/cli/src/program.ts
Andarist: Ah, i guess u are resolving the glob with chokidar, but we could simply just read all filenames in that dir and skip over that.

mattpocock: Yeah I'm trying to test as much of the production code as possible. I found it hard to test the CLI itself (probably missing some knowledge about how to test commander apps).

> This is a little bit iffy for me, especially given that only one specific branch here returns a promise.

Ah, you're right. I'll have a think.

mattpocock:
True, I'll try and find an implementation that doesn't create such a big diff and that stays truer to the implementation.

apps/cli/src/program.ts
Andarist: > Yeah I'm trying to test as much of the production code as possible. I found it hard to test the CLI itself (probably missing some knowledge about how to test commander apps).

Hm, couldn't u `execSync` the compiled CLI? 🤔

mattpocock: Yeah that's what I tried, will try again.

mattpocock:
@Andarist PR much easier to read now!

apps/cli/src/bin.ts
```diff
@@ -46,9 +46,9 @@ const writeToFiles = async (uriArray: string[]) => {
         );
 
         const fileEdits: FileEdit[] = [];
+        let machineIndex = 0;
```
mattpocock, line 49: This was the culprit! Declared the index _inside_ the for loop, doh.

Andarist (approved):

PR #113, opened 2022-03-18
Added quickstart section to docs

mattpocock:


apps/extension/client/README.md
```diff
@@ -1,8 +1,9 @@
 XState VSCode enhances the XState development experience by providing VSCode users with features like visual editing, autocomplete, typegen and linting.
 
-## Installation
+## Quickstart
 
-**[Install via the Visual Studio Code Marketplace →](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode)**
+1. [Install via the Visual Studio Code Marketplace →](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode)
+2. Once installed, you can run `XState: Open Visual Editor` from the command palette to open any machine at your cursor's location.
```
Andarist, line 6: should we also call out the codelens thing somewhere here and mention how one can enable it if they have it disabled by default?

apps/extension/client/README.md
```diff
@@ -4,6 +4,7 @@ XState VSCode enhances the XState development experience by providing VSCode use
 
 1. [Install via the Visual Studio Code Marketplace →](https://marketplace.visualstudio.com/items?itemName=statelyai.stately-vscode)
 2. Once installed, you can run `XState: Open Visual Editor` from the command palette to open any machine at your cursor's location.
+3. If you have code lens enabled, you'll also see 'Open Visual Editor' floating above each `createMachine` call.
```
Andarist, line 7:
```suggestion
3. If you have code lens enabled (this can be enabled using [`editor.codeLens`](https://code.visualstudio.com/docs/getstarted/settings#_default-settings) setting), you'll also see 'Open Visual Editor' floating above each `createMachine` call.
```

Andarist (approved):

apps/extension/client/README.md
davidlav: I would.  The major (game changing?) difference I found between opening the Inspector/Visualizer (ie not the Visual Editor) via the command palette vs via code lens was that when opened with the command palette, it _won't_ auto-update as you make changes to the machine config object.  Only if you open via code lens will it do that, and before stumbled on the issue here that led me down the code lens route, I almost stopped using the extension altogether as the major benefit to having it in-editor is live updating.

Andarist: The callout has already been added in response to my comment, [here](https://github.com/statelyai/xstate-tools/pull/113/commits/c0159b460b90f5b493f3a93e8055a38d41a4d1d9). As to the behavior not being the same - I would say that it's a bug. Could you open a new issue about it?

PR #114, opened 2022-03-18
[Delay merge] Added exit action awareness to typegen algorithm

mattpocock:


packages/shared/src/__tests__/__examples__/exitActions.ts
```diff
@@ -0,0 +1,34 @@
...
+            target: "b",
+          },
+        },
+      },
+      b: {},
+    },
+  },
+  {
+    actions: {
+      doSomethingWithFoo: (context, event) => {
```
Andarist, line 24: I often add an extra event to declared ones and verify that it actually has not been provided here. Cause right now this would kinda pass if the computed type would be `TEvent | { type: 'xstate.stop' }`

packages/shared/src/introspectMachine.ts
```diff
@@ -272,6 +276,10 @@ export const introspectMachine = (machine: XState.StateNode) => {
         actions.addEventToItem(action.type, source, node.path);
       });
     });
+
+    node.onExit.forEach((action) => {
+      actions.addEventToItem(action.type, "xstate.stop", node.path);
```
Andarist, line 281: Since we are adding `xstate.init` to `internalEvents` of the generated typegen info, we should do the same for `xstate.stop`.

.changeset/brave-seas-impress.md
```diff
@@ -0,0 +1,7 @@
+---
+"@xstate/cli": minor
+"@xstate/tools-shared": minor
+"stately-vscode": minor
+---
+
+Gave typegen the ability to calculate all of the events which can fire exit actions.
```
Andarist, line 7: At the moment this will not be type-safe at all. Do we want to wait for landing [the PR in the core](https://github.com/statelyai/xstate/pull/3126) before landing this PR here? I feel like probably a coordinated release would be better. With a callout that this will require a new version of XState to work properly. 

I will also be working next week on "predictable event types for implementations" and we could probably batch releasing it with this change here.

packages/shared/src/introspectMachine.ts
mattpocock: It does that already - anything with `xstate.*` is added.

.changeset/brave-seas-impress.md
mattpocock: Yes, I can dig that. It's a relatively small change this end so we should co-ordinate the release together.

packages/shared/src/introspectMachine.ts
Andarist: ah, got it - didn't know that is automated this way :p

packages/shared/src/introspectMachine.ts
```diff
@@ -188,6 +188,10 @@ export const introspectMachine = (machine: XState.StateNode) => {
     });
 
     node.transitions?.forEach((transition) => {
+      node.onExit.forEach((action) => {
+        actions.addEventToItem(action.type, transition.eventType, node.path);
+      });
```
Andarist, line 193: Hm, I'm not entirely sure but this one seems to be a little bit simplistic. I didn't test it but if I read correctly how this works then it will not work properly for cases like this:
```js
createMachine({
  states: {
    a: {
      exit: 'myAction', // this should receive FOO | 'xstate.stop'
      states: {
        b: {
          on: {
            FOO: '#someIdElsewhere'
          }
        }
      }
    }
  }
})
```

Andarist: The algorithm should account for targets within the subtree being outside of the "source" state that has the `exit` handler defined on it. We should also recheck if this correctly works with internal/external transitions.

mattpocock: 👍  Good point

mattpocock: @Andarist Added a test case to cover this

Andarist: An additional test case that also probably fails is this:
```js
createMachine({
  states: {
    a: {
      on: {
	    FOO: '#someIdElsewhere'
	  },
      states: {
        b: {
          exit: 'myAction', // this should receive FOO | 'xstate.stop'
        }
      }
    }
  }
})
```

mattpocock: Nice, good catch. Fixed

Andarist:
This should also add source events that lead to entering final states to the appropriate exit handlers. Basically, `xstate.stop` is reserved for exit actions called when a machine gets **stopped** from the outside (so any exit action can be called with that as the timing of the `.stop()` call is arbitrary) - but when a machine reaches its final state then exit handlers are called with the last received event.
```js
createMachine({
  initial: 'a',
  states: {
    a: {
      on: { NEXT: 'b' }
    },
    b: {
      type: 'final'
    },
  },
  exit: (ctx, ev) => {
    ev.type // 'xstate.stop' | 'NEXT'
  }
})
```

PR #121, opened 2022-03-21
Fixed an issue with the CLI reporting zero exit code in case of an error

Andarist:


mattpocock (approved):

mattpocock:
@Andarist Could we add a test for this?

Andarist:
I guess that we'll add a test later :P

PR #122, opened 2022-03-21
Fixed an issue where warnings for meta properties would still show even if the user had turned off visual editor warnings

mattpocock:


Andarist (approved):

Andarist:
What's the scenario for disabling those diagnostics? I mean - they are there for a reason, how the user can reliably use the extension if they turn this off? Is it for users using solely the visualizer why not caring about the editor?

mattpocock:
@Andarist Yes, or the typegen and the viz but no editor

PR #126, opened 2022-03-31
Fix parser to include description from transition object

ericjonathan6:
Fix transition parser to include description field.

Currently, when we give description to a transition, it won't be parsed and shown in the editor.
This PR will fix the issue like the screenshot below:

Screenshot 2022-03-31 at 3 13 55 PM

### DIFF
--- a/packages/machine-extractor/examples/transition-description.machine.ts
+++ b/packages/machine-extractor/examples/transition-description.machine.ts
@@ -1,30 +1,30 @@
 import { createMachine } from "xstate";
 
 export const transitionDescription = createMachine({
-	id: "TransitionDescriptionMachine",
-	type: "parallel",
-	states: {
-		Ticket: {
-			initial: "Open",
-			states: {
-				Open: {
-					on: {
-						cancel: {
-							actions: "cancel ticket",
-							target: "Canceled",
-							description: "This is description example",
-						},
-						secondCancel: {
+  id: "TransitionDescriptionMachine",
+  type: "parallel",
+  states: {
+    Ticket: {
+      initial: "Open",
+      states: {
+        Open: {
+          on: {
+            cancel: {
+              actions: "cancel ticket",
+              target: "Canceled",
+              description: "This is description example",
+            },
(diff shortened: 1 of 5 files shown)


mattpocock:
@ericjonathan6 Fantastic! Could you add a test for this? You can do this by adding a machine in the `examples` folder of the machine-extractor package.

packages/machine-extractor/examples/transition-description.machine.ts
```diff
@@ -0,0 +1,23 @@
...
+							target: "Canceled",
+							description: "This is description example",
+						},
+					},
+				},
+				Canceled: {},
+			},
+		},
+	},
+});
```
davidkpiano, line 23: Converted to use spaces for formatting:

```suggestion
export const transitionDescription = createMachine({
  id: "TransitionDescriptionMachine",
  type: "parallel",
  states: {
    Ticket: {
      initial: "Open",
      states: {
        Open: {
          on: {
            cancel: {
              actions: "cancel ticket",
              target: "Canceled",
              description: "This is description example",
            },
          },
        },
        Canceled: {},
      },
    },
  },
});
```

ericjonathan6, line 23: @mattpocock hi, i've added the simple machine for example to show description inside transition, is this enough?

packages/machine-extractor/src/transitions.ts
```diff
@@ -22,6 +22,7 @@ const TransitionObject = objectTypeWithKnownKeys({
   target: TransitionTarget,
   actions: MaybeArrayOfActions,
   cond: Cond,
+  description: StringLiteral,
```
Andarist, line 25: This probably should be:
```suggestion
  description: unionType([StringLiteral, TemplateLiteral]),
```

and we'd need to add a test for that TemplateLiteral case

mattpocock (approved):

mattpocock:
Thanks so much for your contribution!

PR #129, opened 2022-04-05
Fixed a bug where children of states entered from transitions would not have their entry actions or invocations typed properly

### LINKED ISSUE #124: Bug:   eventsCausingServices in *.typegen.ts does not type parallel types correctly
### Description

When using parallel states, service signatures are incorrectly typed. 

eventsCausingServices have the value 'xstate.init'.

repro:
See https://codesandbox.io/s/quirky-poincare-wlofi4

### Expected result

```
eventsCausingServices: {
    off: 'GO_NUMBER' | 'GO_STRING';
    jump: 'GO_STRING';
  };
```

### Actual result

```
eventsCausingServices: {
    off: 'GO_NUMBER';
    jump: 'xstate.init';
  };
```
TypeScript Error:
```
This condition will always return 'false' since the types '"xstate.init"' and '"GO_STRING"' have no overlap.
```

### Reproduction

https://codesandbox.io/s/quirky-poincare-wlofi4

### Additional context

Sorry I couldn't get codesandbox to work with typegen.

mattpocock:
Fixes #124


mattpocock:
Fixes #124

Andarist:
@mattpocock to make GitHub auto-close the referenced issue you must put this "fixes #XXX" comment in the description of the PR itself, it doesn't work if you use this in some latter comment

mattpocock:
Gotcha, it works that way in Linear so I'm getting confused

packages/shared/src/__tests__/__examples__/parallel.ts
```diff
@@ -0,0 +1,71 @@
...
+    initial: "idle",
+    states: {
+      idle: {
+        on: {
+          GO_STRING: {
+            target: "doesnt",
+          },
+          GO_NUMBER: {
+            target: "works",
+          },
```
Andarist, line 30: everything related to this transition is redundant here and somewhat makes analyzing the intent of this test harder

packages/shared/src/__tests__/__examples__/parallel.ts
```diff
@@ -0,0 +1,71 @@
+import { createMachine } from "xstate";
```
Andarist, line 1: since you have added fixes for both invokes and entry actions - it would be great to have dedicated tests for both of those in combination with parallel states. At the moment only invoke case got its test here.

packages/shared/src/introspectMachine.ts
```diff
@@ -191,31 +191,40 @@ export const introspectMachine = (machine: XState.StateNode) => {
...
-              transition.eventType,
-              node.path,
-            );
+          /**
+           * We gather info on all the invocations that begin on
+           * all the nodes that are initial state nodes of the children
+           * of this node (see recursive-invoke)
+           */
+
+          const nodesToGather = [targetNode, ...targetNode.initialStateNodes];
```
Andarist, line 214: note: the logic in this file doesn't handle history nodes at all - so that should be fixed as well, but in a separate PR

packages/shared/src/__tests__/__examples__/parallel.ts
mattpocock: I'll remove this test, this was copied from a user's submission.

Andarist: It was a good test though (if it would be slimmed down) :p resolving compound states recursively and resolving all child states of a parallel state is different. Right now they are both kinda resolved automatically using the same internal thing but we can't guarantee that this will always be the case. So it would be better to have separate tests here for both cases

mattpocock: Re-added

Andarist (approved):

PR #145, opened 2022-05-23
Add vue the languages we detect

mellson:
I think the extension should work equally well in vue files? So I added this to the languages we detect.

### DIFF
--- a/.changeset/fluffy-shoes-judge.md
+++ b/.changeset/fluffy-shoes-judge.md
@@ -1,0 +1,5 @@
+---
+"stately-vscode": patch
+---
+
+added vue to the supported languages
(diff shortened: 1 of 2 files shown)


davidkpiano (approved):

mellson:
> In theory, this might need some testing.

Excellent point, I'll do some tests, and if it works as intended, merge this PR.

mellson:
> In theory, this might need some testing.

Your theory turned out to be right. Our `parseMachinesFromFile` doesn't work on `vue` files.
We're using [babel](https://babeljs.io/docs/en/babel-parser) to parse file contents, which currently doesn't support parsing `vue`files.

Closing, for now, thanks for the input!

gyhyfj:
so sad

Andarist:
@gyhyfj we'll be adding support for Vue and Svelte in the coming months

gyhyfj:
Thank you very much !

matjazPrijatelj:
Any update on this since or still the only option to use this in vue files is to split xstate in seperate js file?

PR #146, opened 2022-05-23
Add file nesting to typegen files

mellson:
This adds file nesting for typegen files if the user has file nesting enabled. Thanks to Erik Rasmussen for the tip!


Andarist (commented):
1. I wonder if we could smartly enable this by default or make it easier for people to enable it? It should be possible to update config using the vscode API: https://code.visualstudio.com/api/references/vscode-api#1513
2. could we add a recommendation/note about this to the README?

mattpocock:
@mellson I'm also nervous about having this as a default because it makes the behaviour less discoverable for users.

But we could enable it as a config option _ourselves_, and use our config option to update VSCode's config? Many possibilities.

mellson:
Thank you both for the input. I have now updated this with our own config option: `xstate.nestTypegenFiles`. This is set to `true` by default.

When launching VSCode we will check if that config is `true` if it is we will check if there's already a pattern for nesting typegen files, if not we show a little prompt to help the user enable or dismiss it.

CleanShot 2022-05-23 at 13 57 07

mellson:
2 mins, just discovered a possible improvement when updating config files

mellson:
Ready for review 👍🏻

apps/extension/client/src/extension.ts
```diff
@@ -63,20 +64,22 @@ export async function activate(context: vscode.ExtensionContext) {
...
     serverOptions,
-    clientOptions,
+    clientOptions
   );
 
   client.start();
 
   await client.onReady();
 
+  checkTypegenNestingConfiguration();
```
Andarist, line 74: I assume that this will also get executed right after the installation as this is called in the `activate` handler, right?

Andarist:
`.DS_Store` file has been accidentally committed here - we should remove it and add it to the `.gitignore` patterns

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,57 @@
...
+const hasTypeGenPattern = (fileNestingPatterns: object) =>
+  Object.values(fileNestingPatterns).some((pattern: string) => {
+    if (pattern.includes(",")) {
+      const patterns = pattern.split(",");
+      return patterns.some(
+        (innerPattern) => innerPattern.trim() === typeGenPattern
+      );
+    } else {
+      return pattern === typeGenPattern;
+    }
```
Andarist, line 14: nit: for simplicity of the check I would probably go with this:
```ts
    const patterns = pattern.split(",");
    return patterns.some(
      (innerPattern) => innerPattern.trim() === typeGenPattern
    );
```
It should behave in the same way

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+export const checkTypegenNestingConfiguration = () => {
+  const fileNestingConfig = vscode.workspace.getConfiguration(
+    "explorer.fileNesting"
+  );
```
Andarist, line 20: I wonder what this would return in pre-1.67. There is a chance that this might be undefined? or some empty obj? It would be nice if we'd make sure that we don't end up crashing in older VScode versions

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,57 @@
+import * as vscode from "vscode";
+
+const typeGenPatternKey = "*.ts";
+const typeGenPattern = "${capture}.typegen.ts";
+const hasTypeGenPattern = (fileNestingPatterns: object) =>
```
Andarist, line 5: we are currently only checking workspace settings, I wonder if we should respect user setting here too - although I don't have strong recommendations as to how this should work. Pehaps we should check both for the pattern but keep writing to workspace-level settings?

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+      return pattern === typeGenPattern;
+    }
+  });
+
+export const checkTypegenNestingConfiguration = () => {
+  const fileNestingConfig = vscode.workspace.getConfiguration(
+    "explorer.fileNesting"
+  );
+  const fileNestingPatterns = fileNestingConfig.get<object>("patterns");
```
Andarist, line 21: q: does this contain defaults from here?
https://github.com/microsoft/vscode/blob/ecf7f9cfd9ab5460cc2203fb1eb56ae2569157e6/src/vs/workbench/contrib/files/browser/files.contribution.ts#L503-L510

apps/extension/client/src/checkTypegenNesting.ts
mellson: I like it

apps/extension/client/src/checkTypegenNesting.ts
mellson: we check the full config now since we're not providing a scope when we get the configuration.

apps/extension/client/src/checkTypegenNesting.ts
mellson: We include all the patterns the user already has in there. And it's most likely the defaults, so we keep them and add our needed pattern to the patterns for typescript.

mellson:
Sorry, just converting this to a draft since I just realized we need a mechanism for disabling the settings again if a user uses our config option to disable nesting.

Andarist:
> Sorry, converting this to a draft since I just realized we need a mechanism for disabling the settings again if a user uses our config option to disable nesting.

could u describe the scenario in more detail? Im not sure if i understand it

Andarist:
> 2 - The user opens VSCode with xstate.nestTypegenFiles set to false. But there is a pattern nesting the files. We remove our pattern, so the files don't nest in this scenario. This can happen if they update their config before our extension is fully loaded, or they update the config outside VSCode.

At first, us **removing** a pattern that could have been configured manually seemed off but I guess this shouldn't be harmful anyhow since the nesting setting in XState's extension will be on by default 🤔

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,150 @@
+import * as vscode from "vscode";
+import { WorkspaceConfiguration } from "vscode";
+
+const typeGenPatternKey = "*.ts";
```
Andarist, line 4: We also need to handle `.tsx` files. And TS 4.7 will also come with two new extensions - `.mts` and `.cts`. We don't necessarily support those 2 new extensions anywhere else right now so it's OK to skip them but perhaps we should add a TODO comment about adding support for them in the future or smth.

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  // If there's no patterns defined for ts files we don't need to do anything
+  if (existingTsPattern === "") return;
+
+  // Remove our typegen pattern from the existing patterns
+  const updatedPattern = existingTsPattern
+    .split(",")
+    .map((part) => part.trim())
+    .filter((part) => part !== typeGenPattern)
+    .join(",");
```
Andarist, line 60: nit: you are removing whitespace from the user's config - ideally we should preserve the original formatting

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,150 @@
...
+    true
+  );
+};
+
+const disableTypeGenNesting = (
+  xstateConfig: WorkspaceConfiguration,
+  fileNestingConfig: WorkspaceConfiguration,
+  fileNestingPatterns: object
+) => {
+  xstateConfig.update("nestTypegenFiles", false, true);
```
Andarist, line 47: I am somewhat confused by this line - especially when it's called from within `onDidChangeConfiguration`. If a configuration just has been changed - why do we need to update it again here? I see the third argument here that makes it to change the global config, but what's that? 🤔

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,150 @@
...
+    vscode.window
+      .showInformationMessage(
+        "Do you want to enable file nesting for XState typegen files?",
+        enableOption,
+        disableOption
+      )
+      .then((choice) => {
+        switch (choice) {
+          case enableOption:
+            enableTypeGenNesting(fileNestingConfig, fileNestingPatterns);
```
Andarist, line 135: nit: there is a small race condition here - there is some time between we read (and cached) those variables that are used as arguments here and between the time we get to use them. Often when dealing with this kind of a logic I've seen people to defensively re-obtain the **current** config items, just to make sure that such races are minimized.

Perhaps both `enableTypeGenNesting` and `disableTypeGenNesting` should just read those on their own and shouldn't receive those values as arguments? You could create a helper for reading normalized & typed values in a single call.

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,150 @@
...
+  // VSCode's configuration for file nesting
+  const fileNestingConfig = vscode.workspace.getConfiguration(
+    "explorer.fileNesting"
+  );
+
+  // If there's no fileNestingConfig with patterns then we can't check the nesting, maybe the user is on a version of VSCode before 1.67
+  if (!fileNestingConfig || !fileNestingConfig.has("patterns")) return;
+
+  // VSCode's configuration for xstate
+  const fileNestingPatterns = fileNestingConfig.get<object>("patterns");
```
Andarist, line 115: How safe this generic param is? Can I do something stupid like editing `patterns` manually to a `null` or smth and blow up our extension this way? :P Or is there some kind of a JSON schema applied on VScode's side that prevents such errors?

apps/extension/client/src/checkTypegenNesting.ts
mellson: Great input, I've added all four filetypes.

apps/extension/client/src/checkTypegenNesting.ts
mellson: Good catch, I've updated this to `...join(", ")` to keep the default vscode formating. OK?

apps/extension/client/src/checkTypegenNesting.ts
mellson: This was mainly done so that we could use the same helper in the three places it's called. But our call to `xstateConfig.update("nestTypegenFiles", false, true)` was actually just useful in one of those places, so I moved it there.

apps/extension/client/src/checkTypegenNesting.ts
mellson: Agree that would be better, updated in the latest commit

apps/extension/client/src/checkTypegenNesting.ts
mellson: VSCode does have a schema, but it's not enforced, so I added some type guards. However, we can't guard completely, so if a user does `"explorer.fileNesting.patterns": null` in their settings, we will get an error because this call `vscode.workspace.getConfiguration("explorer.fileNesting")` will fail.

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,150 @@
...
+// Checks if the user has our typegen pattern set in their workspace config
+const hasTypeGenPattern = (fileNestingPatterns: object) =>
+  Object.values(fileNestingPatterns).some((pattern: string) =>
+    pattern
+      .split(",")
+      .map((part) => part.trim())
+      .some((part) => part === typeGenPattern)
+  );
+
+const enableTypeGenNesting = (
```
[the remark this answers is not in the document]
mellson: 👍🏻

apps/extension/client/src/checkTypegenNesting.ts
mellson: Regarding the third argument to `update`, setting this to true will update the global settings (in my case this is my synced user settings). If this is false it will just update the workspace settings. My take is that this config is probably something a user wants to set globally as the default, if not they can still manually set their workspace settings.

apps/extension/client/src/checkTypegenNesting.ts
Andarist: It's probably better as it matches the default formatting now - but a nitpick would be that it still doesn't preserve the input formatting. A version that would do this could look smth like this:
```js
  const updatedPattern = existingTsPattern
    .split(",")
    .filter((part) => {
        const trimmed = part.trim();
        return trimmed !== typeGenPattern;
    })
    .join(",");
```

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,166 @@
...
+
+const typegenPatternKeys = ["*.ts", "*.tsx", "*.mts", "*.cts"];
+const typegenPattern = "${capture}.typegen.ts";
+
+const getXStateConfig = () => vscode.workspace.getConfiguration("xstate");
+const getFileNestingConfig = () =>
+  vscode.workspace.getConfiguration("explorer.fileNesting");
+
+// Checks if the user has our typegen pattern set in their workspace config
+const hasTypeGenPattern = (fileNestingPatterns: object) =>
```
Andarist, line 11:
```suggestion
const hasTypegenPattern = (fileNestingPatterns: object) =>
```

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,166 @@
...
+
+const createTypeGenPatterns = (transformer: (patternKey: string) => string) => {
+  const transformedArray = typegenPatternKeys.map((patternKey) => ({
+    [patternKey]: transformer(patternKey),
+  }));
+  // Return an object with all the transformed keys
+  return transformedArray.reduce(function (result, item) {
+    var key = Object.keys(item)[0]; //first property: a, b, c
+    result[key] = item[key];
+    return result;
```
Andarist, line 27:
```suggestion
  return transformedArray.reduce(function (result, item) {
    const firstKey = Object.keys(item)[0];
    result[firstKey] = item[firstKey];
    return result;
```

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,166 @@
...
+  const fileNestingPatterns = fileNestingConfig.get("patterns");
+  if (typeof fileNestingPatterns !== "object") return;
+
+  const getUpdatedPattern = (typegenPatternKey: string) => {
+    // Check if the user has a pattern defined for ts files
+    const existingTsPattern: string | undefined =
+      fileNestingPatterns[typegenPatternKey];
+
+    // If there's no patterns defined for ts files we don't need to do anything
+    if (!existingTsPattern || existingTsPattern === "") return undefined;
```
Andarist, line 80: If I read the code correctly then according to the declared signature of `createTypeGenPatterns` this should be a type error because the callback type there is typed as `(patternKey: string) => string`

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,166 @@
...
+      ...fileNestingPatterns, // Reuse existing patterns
+      ...updatedPatterns, // Add our updated patterns
+    },
+    true
+  );
+};
+
+const removeEmptyPatterns = (obj: object) =>
+  Object.fromEntries(
+    Object.entries(obj).filter(([_, pattern]) => pattern != "")
```
Andarist, line 65: the non-strict eq sign here might be problematic when reading this code - I've missed it at first and I've thought briefly that there is a logic error here (or at the consuming call site). I would prefer if we'd stick to `===` and `!==`

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,166 @@
+import * as vscode from "vscode";
+
+const typegenPatternKeys = ["*.ts", "*.tsx", "*.mts", "*.cts"];
```
Andarist, line 3: q: are there any docs for what type of patterns are actually supported here by VSCode? I wonder if any other, more complex, patterns are supported 🤔

apps/extension/client/src/checkTypegenNesting.ts
mellson: Thanks, that's an error - I just missed it

apps/extension/client/src/checkTypegenNesting.ts
mellson: I couldn't find any docs on this, I tried adding patterns like `*.ts*` but this was rejected by their schema

Andarist: I think the patterns are pretty simple and this should work OK.

We can take a look at the markdown description here:
https://github.com/microsoft/vscode/blob/1e60a4400f4fa00f65c6e212362153ec58716871/src/vs/workbench/contrib/files/browser/files.contribution.ts#L488

and valid substation types here:
https://github.com/microsoft/vscode/blob/d9225482faa9922efd0e88058c1bb9152d17a615/src/vs/workbench/contrib/files/common/explorerFileNestingTrie.ts#L224-L229

and patterns for keys are somewhat checked/processed here:
https://github.com/microsoft/vscode/blob/d9225482faa9922efd0e88058c1bb9152d17a615/src/vs/workbench/contrib/files/common/explorerFileNestingTrie.ts#L166-L185

This doesn't show the full picture but I think that basically only a single wildcard is allowed, other than that you can use an exact matches. So the current implementation looks good as no other "composite" patterns should be allowed.

apps/extension/client/src/checkTypegenNesting.ts
```diff
@@ -0,0 +1,166 @@
...
+// Checks if the user has our typegen pattern set in their workspace config
+const hasTypegenPattern = (fileNestingPatterns: object) =>
+  Object.values(fileNestingPatterns).some((pattern: string) =>
+    pattern
+      .split(",")
+      .map((part) => part.trim())
+      .some((part) => part === typegenPattern)
+  );
+
+const createTypeGenPatterns = (transformer: (patternKey: string) => string) => {
```
Andarist, line 19:
```suggestion
const createTypegenPatterns = (transformer: (patternKey: string) => string) => {
```

Andarist (approved):

apps/extension/client/src/checkTypegenNesting.ts
mellson: Wow, nice finds - thank you!

PR #150, opened 2022-05-30
Open documentation based on where you are in a machine

mellson:
Seems to work ok, but need to discuss the approach before commiting to it.

### DIFF
--- a/apps/extension/client/src/documentationLinks.ts
+++ b/apps/extension/client/src/documentationLinks.ts
@@ -1,0 +1,51 @@
+const baseUrl = "https://xstate.js.org/docs/guides";
+
+interface DocumentationLink {
+  keyWords: string[];
+  description: string;
+  link: string;
+}
+
+const documentationLinks: DocumentationLink[] = [
+  {
+    keyWords: ["actions"],
+    description: "Actions",
+    link: `${baseUrl}/actions.html`,
+  },
+  {
+    keyWords: ["context"],
+    description: "Context",
+    link: `${baseUrl}/context.html`,
+  },
+  {
+    keyWords: ["initial"],
+    description: "Initial Context",
+    link: `${baseUrl}/context.html#initial-context`,
+  },
+  {
+    keyWords: ["events"],
+    description: "Events",
+    link: `${baseUrl}/events.html`,
+  },
+  {
+    keyWords: ["states"],
+    description: "States",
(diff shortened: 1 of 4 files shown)


mellson:
Hi @mattpocock, this is now ready for review. After talking to Laura today, we decided to point this to the new docs. I've added a non-exhaustive list of keywords, but enough to be helpful, I think.

But it just hit me that we could also add the links to the types we ship with XState. What's your take? Is this still something we want here, or should we close this in favor of adding links to the XState types?

apps/extension/client/src/documentationLinks.ts
```diff
@@ -1,46 +1,72 @@
...
+    link: `${baseUrl}/actions/context#updating-context`,
   },
   {
     keyWords: ["tsTypes", "typegen"],
     description: "Typegen",
-    link: `${baseUrl}/typescript.html#typegen`,
+    link: `${baseUrl}/typescript/typegen`,
   },
   {
     keyWords: ["schema"],
```
mattpocock, line 52: `schema` is tricky because there's no central spot where we talk about how to use - we split it up over several files. So what you want is to ensure we're in `schema.context` to go to the context section, `schema.events` to go to the events section, and `schema.services` to talk about invoked services with typegen.

apps/extension/client/src/documentationLinks.ts
```diff
@@ -1,46 +1,72 @@
...
+    description: "Guards",
+    link: `${baseUrl}/transitions-and-choices/guards`,
+  },
+  {
+    keyWords: ["createMachine"],
+    description: "Running machines",
+    link: `${baseUrl}/running-machines/intro`,
+  },
+  {
+    keyWords: ["createTestMachine"],
```
mattpocock, line 67: `createTestModel` should probs be here too

apps/extension/client/src/documentationLinks.ts
mellson: Thank you for the feedback; I have added some special handling for the schema object in the latest commit. But I need to figure out where to link to in the three cases:

1. schema.context -> https://graph-docs.vercel.app/actions/context#type-script
2. schema.events -> ?
3. schema.services -> ?

Can you help with the correct links?

mellson:
I am closing this in favor of adding the links to XState's JSDoc (at some point).

PR #151, opened 2022-05-30
Added extraction for createTestMachine

mattpocock:
This is important because we're adding `createTestMachine` as a new API and need to make sure the VSCode extension works with it.


mellson (approved):
Looks good 👍🏻

PR #156, opened 2022-06-07
Clicking Give Feedback option gives a blank page in VSC extension

mellson:
This is the extension part of this task, there's another part in [the studio repo](https://github.com/statelyai/studio/pull/625) to allow for sending messages

### DIFF
--- a/apps/extension/client/src/editorWebviewScript.ts
+++ b/apps/extension/client/src/editorWebviewScript.ts
@@ -152,6 +152,7 @@ const machine = createMachine<WebViewMachineContext, EditorWebviewScriptEvent>(
 
             if (!iframe || iframe.src) return;
 
+            // Here we add the receiver for messages posted by the editor web app
             window.addEventListener(
               "message",
               (e) => {
--- a/apps/extension/client/src/editorWebviewScript.ts
+++ b/apps/extension/client/src/editorWebviewScript.ts
@@ -1,10 +1,7 @@
-import { assign } from "xstate";
-import { MachineConfig } from "xstate";
-import { interpret } from "xstate";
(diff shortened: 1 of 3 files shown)


apps/extension/client/src/initiateEditor.ts
```diff
@@ -44,20 +44,20 @@ export const initiateEditor = (
...
       return;
     } else if (result === "timed-out") {
       vscode.window.showErrorMessage(
-        "The authentication request timed out. Please try again.",
+        "The authentication request timed out. Please try again."
       );
       return;
     } else if (result === "unknown-error") {
       vscode.window.showErrorMessage(
-        "An unknown error occurred. Please try again.",
```
mellson, line 60: Note to self: add a prettier config to the tools repo 🎨

apps/extension/client/src/editorWebviewScript.ts
```diff
@@ -168,9 +171,13 @@ const machine = createMachine<WebViewMachineContext, EditorWebviewScriptEvent>(
               false
             );
 
+            /**
+             * We add runningInStatelyExtension to the iframe's src.
+             * This is used by `useVsCode` in the editor to check that the extension is running the web app.
+             * */
```
Andarist, line 177: I'm not 100% sure as this would have to be tested but I think that you might already have this information in the Studio and that you don't need to pass it explicitly from here. You could try checking `typeof acquireVsCodeApi !== 'undefined'` there - I believe that it should be a function in VSCode's webviews and it ain't anywhere else :p

Andarist (approved):
This needs a short/somewhat vague changeset for the change to be released after merging

mattpocock (approved):

PR #157, opened 2022-06-08
Added shared settings for prettier and vscode

mellson:
Adding some default shared settings. This will help us achieve similar formatting of the files, and that will help highlight the relevant changes in future code reviews. I didn't include a changeset since this is an internal change.

Should we run a prettier format on all files in the repo before merging this?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,11 @@
 {
-	"editor.insertSpaces": false,
-	"tslint.enable": true,
-	"typescript.tsc.autoDetect": "off",
-	"typescript.preferences.quoteStyle": "single",
-	"editor.codeActionsOnSave": {
-		"source.fixAll.eslint": true
-	}
+  "editor.insertSpaces": false,
+  "tslint.enable": true,
+  "typescript.tsc.autoDetect": "off",
+  "typescript.preferences.quoteStyle": "single",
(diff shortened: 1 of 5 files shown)


.vscode/settings.json
```diff
@@ -1,9 +1,11 @@
...
-	"tslint.enable": true,
-	"typescript.tsc.autoDetect": "off",
-	"typescript.preferences.quoteStyle": "single",
-	"editor.codeActionsOnSave": {
-		"source.fixAll.eslint": true
-	}
-}
\ No newline at end of file
+  "editor.insertSpaces": false,
+  "tslint.enable": true,
```
Andarist, line 3: We are not using TSLint afaik so this probably should be removed

.vscode/settings.json
```diff
@@ -1,9 +1,11 @@
...
+  "typescript.tsc.autoDetect": "off",
+  "typescript.preferences.quoteStyle": "single",
```
Andarist, line 5: does it have an effect of anything if we are using prettier anyway?

mellson: not sure, I just left it there since it was already in there. But I'll remove it now that we have the prettier config, which should handle it

Andarist (approved):

PR #160, opened 2022-06-08
add `stateValues` to typegen output

brunocrosier:
- Fix issue where generated `.typegen.ts` files in `__examples__` folder would not be deleted 
- Add `stateValues` to typegen output, so that it can be used in https://github.com/statelyai/xstate/pull/3342


Andarist:
I know that it has been a while since you have created this PR - but could you rebase this on top of the main branch and add snapshots tests for the `getTypegenOutput` (the test setup has been refactored in the meantime)? I'd like to see how the generated output looks like to decide if we should include this as part of the typegen types or not.

packages/shared/src/getTransitionsFromNode.ts
```diff
@@ -1,89 +1,107 @@
...
+    machine.getStateNodeById(id)
+  );
 
-	const states = allStateNodes.reduce((arr: string[], node) => {
-		return [
-			...arr,
-			...toStatePaths(pathToStateValue(node.path)).map((path) => path.join('.')),
-		];
-	}, [] as string[]);
+  const states = allStateNodes.reduce((arr: StateValue[], node) => {
```
brunocrosier, line 93: use `StateValue` type 

every other change in this file is just formatting

brunocrosier:
> I know that it has been a while since you have created this PR - but could you rebase this on top of the main branch and add snapshots tests for the `getTypegenOutput` (the test setup has been refactored in the meantime)? I'd like to see how the generated output looks like to decide if we should include this as part of the typegen types or not.

Sorry I lost a bit of context on this since opening the PR but I hope these are the changes you requested 🙂

Andarist:
Yeah, this is what I had in mind - thank you 🙏 I'm gonna try to take a look at this soon

PR #168, opened 2022-06-27
WIP Prototype: getMachine

mellson:
This is a very early prototype that allows the CLI to fetch machines from the Stately Editor based on machine id's.


mellson:
Closing in favor of something else we're working on...

PR #170, opened 2022-06-29
Refactor `introspectMachine`

Andarist:


packages/shared/src/introspectMachine.ts
```diff
@@ -1,41 +1,61 @@
 import { INLINE_IMPLEMENTATION_TYPE } from "@xstate/machine-extractor";
```
mattpocock, line 1: The code here looks like a solid improvement, and if the tests are passing I'm happy.

So I understand the goals:

Refactor this so that it's easier to read, and uses a shared TraversalContext.
Support a couple of use cases that we weren't testing for before, namely exit actions in deep children of states.

Was there anything I missed here?

Andarist: > Support a couple of use cases that we weren't testing for before, namely exit actions in deep children of states.

I think those were already sort of covered in https://github.com/statelyai/xstate-tools/pull/114 but that PR only has covered external transitions whereas exit actions are still called for internal transitions, so I've fixed that. This PR fixed a couple of cases like that, around reentrancy of states and it also follows the **ancestry** chain to add events correctly there too.

I'll try to add more tests in a moment but the improvements are also around some entry actions cause previously we were missing cases like this:
```json
{
  states: {
    a: {
      on: {
        FOO: '#b1'
      }
    }
    b: {
      entry: ['makeSmthWithFoo'], // this should receive FOO
      states: {
        b1: {}
      }
    }
  }
}
```
In a similar manner, events should be received by all ancestors up to the common ancestor of the source and the target. So I've implemented those improved graph traversal here, hopefully replicating the actual logic correctly (I spent quite a bit of time on this, even fixing 2 cases in the XState itself in the process, so hopefully it's alright now). 

There are still some cases that are not handled though (like exit handlers combined with final states) but I will follow up on them with separate PRs.

PR #173, opened 2022-06-30
[cli] Add new options to watcher

RafalFilipek:
# What

- add `--poll` / `-p` option  for polling 
- add `--ignored` / `-i` option for ignoring specified paths 
- add some default options for watcher based on config used in `tailwindcss`

# Why

**`--poll` option**

It was easy to add and maybe someone will need that

**`--ignored` option**

This might be useful for big projects / monorepos. Someone might want to ignore `node_modules` directory. My case was slightly different. Due to unexpected behavior of `chokidar` lib on Windows,  I had to ignore some additional paths. Nevertheless, the case is still valid :)

**Additional options**
This was copied from https://github.com/tailwindlabs/tailwindcss/blob/60a0ae20176604b9444f95b09cbb1f549702a143/src/cli.js#L846-L854. Since `tailwind` is widely used, I assumed those settings might serve as nice defaults.

Thank you for your time.

### DIFF
--- a/apps/cli/src/bin.ts
+++ b/apps/cli/src/bin.ts
@@ -102,37 +102,69 @@ program
   .description("Generate TypeScript types from XState machines")
   .argument("<files>", "The files to target, expressed as a glob pattern")
   .option("-w, --watch", "Run the typegen in watch mode")
-  .action(async (filesPattern: string, opts: { watch?: boolean }) => {
-    if (opts.watch) {
-      // TODO: implement per path queuing to avoid tasks related to the same file from overlapping their execution
-      const processFile = (path: string) => {
-        if (path.endsWith(".typegen.ts")) {
-          return;
-        }
-        writeToFiles([path]).catch(() => {});
-      };
-      // TODO: handle removals
-      watch(filesPattern, { awaitWriteFinish: true })
-        .on("add", processFile)
-        .on("change", processFile);
-    } else {
-      const tasks: Array<Promise<void>> = [];
-      // TODO: could this cleanup outdated typegen files?
-      watch(filesPattern, { persistent: false })
-        .on("add", (path) => {
+  .option(
+    "-p, --poll",
+    "Use polling instead of filesystem events when watching"
+  )
+  .option(
+    "-i, --ignored <patterns...>",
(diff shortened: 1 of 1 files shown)


apps/cli/src/bin.ts
```diff
@@ -102,37 +102,69 @@ program
...
-      watch(filesPattern, { awaitWriteFinish: true })
-        .on("add", processFile)
-        .on("change", processFile);
-    } else {
-      const tasks: Array<Promise<void>> = [];
-      // TODO: could this cleanup outdated typegen files?
-      watch(filesPattern, { persistent: false })
-        .on("add", (path) => {
+  .option(
+    "-p, --poll",
```
Andarist, line 106: are those `-p` and `-i` shortcuts matching the Tailwind CLI? Personally, I would not introduce those as we might add some other options in the future and they might be better suited for having those shortcuts

apps/cli/src/bin.ts
```diff
@@ -102,37 +102,69 @@ program
...
+    "Use polling instead of filesystem events when watching"
```
Andarist, line 107: is there any specific reason to prefer this in certain situations?

apps/cli/src/bin.ts
```diff
@@ -102,37 +102,69 @@ program
...
-          tasks.push(writeToFiles([path]));
+          writeToFiles([path]).catch(() => {});
+        };
+
+        // TODO: handle removals
+        watch(filesPattern, {
+          ignored,
+          usePolling: shouldPoll,
+          interval: shouldPoll ? pollInterval : undefined,
+          ignoreInitial: true,
```
Andarist, line 138: this one has been added, right? could you elaborate on why you think it's needed? what kind of issues this can help us avoid?

apps/cli/src/bin.ts
```diff
@@ -102,37 +102,69 @@ program
...
+          awaitWriteFinish: shouldCoalesceWriteEvents
+            ? {
+                stabilityThreshold: 50,
+                pollInterval: pollInterval,
+              }
+            : false,
```
Andarist, line 144: previously we were using `true` here at all times - any particular reason why this has been changed?

PR #176, opened 2022-07-04
Refactor test setup for `getTypegenOutput`

Andarist:
The main goal here was to make it easier to:
- compare the old and current output - when refactoring this is crucial, it's also useful when just exploring the codebase etc (this has been solved by using snapshots)
- run a single test case, also crucial when developing/refactoring and focusing on isolated failures (this has been done by dropping usage of the `tsc` from the test, creating a test block for each test, and using `it.only` instead of `it` based on the filename pattern)

As mentioned, I've removed `tsc` from those tests - at first, I was creating a more complicated setup. It was somewhat neat as it allowed us to run `tsc` for each test separately but I've concluded that at the end of the day it wasn't worth it and that actually those unit tests here don't have to be concerned with type-checking at all.

I've also fought a good fight with monorepo setup and as a result I can know execute tests with VScode attached to the debugged process.


apps/cli/jest.config.js
```diff
@@ -10,8 +10,9 @@ module.exports = {
   ],
   snapshotFormat: {
     printBasicPrototype: false,
+    escapeString: false,
   },
   transform: {
-    "^.+\\.tsx?$": "esbuild-jest",
+    "^.+\\.tsx?$": ["esbuild-jest", { sourcemap: true }],
```
mellson, line 16: Nice ⚡

babel.config.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
-  presets: ["@babel/preset-typescript"],
+  presets: [
+    ["@babel/preset-env", { targets: { node: "14" } }],
```
mellson, line 3: Is node 14 a suitable target here when we have defined v16 in the nvm file?

mellson (approved):
I only have the one question about the node version - everything else looks good. So feel free to merge when you think it's done.

babel.config.js
Andarist: Those are somewhat 2 different things:
- `.nvmrc` defines a node version that should be used when working in the repository, when working
- this target here defines what is the lowest common denominator for the node version that people can use when installing our packages

mellson: Makes sense, thanks for the answer

PR #182, opened 2022-07-05
Fixed an issue with `initialStateNodes` resolution in `introspectMachine`

Andarist:


packages/shared/src/getTypegenOutput.ts
```diff
@@ -202,20 +202,13 @@ const collectInternalEvents = (lineArrays: { events: string[] }[][]) => {
...
-         * to be caused by xstate.init.
-         */
-        "'xstate.init'"
-      };`;
+      return `${JSON.stringify(line.name)}: ${unique(
+        line.events.map((event) => {
+          return event;
+        })
+      )
+        .map((event) => JSON.stringify(event))
```
Andarist, line 210: we might consider sorting those here as otherwise the generated file gets sensitive to our internal logic - this might be observed in one of the updated snapshots in this PR since the events got merely reordered there

packages/shared/src/getTypegenOutput.ts
```diff
@@ -202,20 +202,13 @@ const collectInternalEvents = (lineArrays: { events: string[] }[][]) => {
...
-          line.events.map((event) => {
-            return event;
-          })
-        )
-          .map((event) => JSON.stringify(event))
-          .join(" | ") ||
-        /**
-         * If no transitions go to this guard/service/action, it's guaranteed
-         * to be caused by xstate.init.
-         */
```
Andarist, line 216: I believe that everything should already be handled upfront and there is no need to have this fallback here - it could only obscure some bugs for us.

packages/shared/src/__tests__/getTypegenOutput.test.ts
```diff
@@ -15,24 +15,27 @@ describe("getTypegenOutput", () => {
...
+      expect(
+        format(getTypegenOutput(event), { parser: "typescript" })
+      ).toMatchSnapshot();
+    };
+
+    const extensionlessFile = file.slice(0, -path.extname(file).length);
+
+    if (/\.only$/.test(extensionlessFile)) {
+      // preserve original test name
+      it.only(extensionlessFile.replace(/\.only$/, ""), runTest);
```
Andarist, line 36: we kinda don't need it since it's now possible to use `testNamePattern` (or friendly filtering with suggestions in the CLI, thanks to `jest-watch-typeahead/testname`) but OTOH I've already added it and might useful occasionally

packages/shared/src/introspectMachine.ts
```diff
@@ -351,7 +351,17 @@ function collectEnterables(ctx: TraversalContext, node: XState.StateNode) {
     return;
   }
 
-  const enterableNodes = [node, ...node.initialStateNodes];
+  const enterableNodes = new Set([node]);
+  const initialLeafNodes = node.initialStateNodes;
```
Andarist, line 355: You can confirm how it behaves here: https://codesandbox.io/s/small-glitter-strroo?file=/src/index.js

packages/shared/src/getTypegenOutput.ts
Andarist: Actually - thanks to the test in the CLI, I've realized that sometimes things might actually be unreachable. But in that case, it's more correct to default to `never` here instead of `'xstate.init'`

mellson (approved):
Love reviewing your code, so clean and plenty of useful comments 👏🏻

PR #188, opened 2022-07-21
Changes made outside VS Code are now sent to the visual editor

mellson:
This PR adds a new setting that allows for changes being sent to the visual editor after every change to a document. We send changes after a save in VS Code; this makes sense and will stay the default. But if a user makes changes outside VS Code and only use Code to show the visual editor this is confusing - https://discord.com/channels/795785288994652170/885197642357960764/997294408678506668

Now we allow the user to set `xstate.sendChanges` to `onCreation`; this will send changes to the editor after each keypress / change. The changes are debounced with a default delay of 1 second; the user can change this with the new setting `xstate.sendChangesDelay`.


apps/extension/client/src/initiateEditor.ts
```diff
@@ -129,25 +126,57 @@ export const initiateEditor = (
...
+
+    if (sendChangesOnCreation) {
+      const debouncedSendChangesToVisualEditor = debounce(
+        sendChangesToVisualEditor,
+        sendChangesDelay
+      );
+
+      return vscode.workspace.onDidChangeTextDocument(({ document }) =>
+        debouncedSendChangesToVisualEditor(document)
+      );
```
Andarist, line 162: I think that this might lead to some weird issues when you break the config temporarily (not quite a new thing that might happen but the frequency of this might get increased here).

I wonder... is there a way to recognize a change of the document like the one described in the linked Discord comment? I've skimmed through the API docs and I don't quite see an obvious way to do that.

I think that **ideally** we'd:
- update immediately if we can recognize a change like the one described in the comment
- have a "smart" live update mechanism that could become a default one day (probably the implemented denounced logic is a good-enough start for that)

apps/extension/client/package.json
```diff
@@ -81,6 +81,20 @@
           "type": "boolean",
           "default": true,
           "description": "Enable nesting the generated typegen files."
+        },
+        "xstate.sendChanges": {
```
Andarist, line 85: `sendChanges` seems to be somewhat close to the implementation - from the user perspective I would call it a refresh or a sync (or maybe even something else)

apps/extension/client/package.json
```diff
@@ -81,6 +81,20 @@
...
+          "type": "string",
+          "enum": [
+            "onSave",
+            "onCreation"
```
Andarist, line 89: this option name seems to be a little bit confusing to me since I don't know what "creation" refers to here - if i understand correctly it's mote like `onTyping`

apps/extension/client/src/initiateEditor.ts
```diff
@@ -129,25 +126,57 @@ export const initiateEditor = (
...
+    } else {
+      return vscode.workspace.onDidSaveTextDocument(sendChangesToVisualEditor);
```
Andarist, line 164: Perhaps this should still result in an immediate refresh of the editor even if the refreshing strategy is set to typing?

apps/extension/client/package.json
farskid: Prettier calls them `onSave`, `onPaste` and `onType`.

apps/extension/client/src/utils.ts
```diff
@@ -0,0 +1,13 @@
+export const debounce = <T extends (...args: any[]) => ReturnType<T>>(
+  callback: T,
+  timeout: number
+): ((...args: Parameters<T>) => void) => {
+  let timer: ReturnType<typeof setTimeout>;
```
farskid, line 5:
```suggestion
  let timer: number;
```
`setTimeout` always returns a number, so it felt weird to use return type.

Andarist: It doesn't return a number in node JS :D It might be easier to grab the correct type here using `ReturnType` like it has been done than to try to remember how to reach for that type manually (it's `NodeJS.Timer` in node)

farskid: TIL!

mellson:
@farskid @Andarist 👋🏻 Thank you both for your review and comments! I've found a more straightforward way to detect changes made outside VS Code and use that as the default.

Now we always use `onDidChangeTextDocument` over `onDidSave`. We can use the `isDirty` flag to determine if it's a file save. The only remaining problem is that `onDidChangeTextDocument` is called multiple times. But that is solved by comparing the document path and text to what we already sent to the editor.

It seems to work fine in my tests. What do you think about this new approach?

farskid (approved):

farskid:
Great work on this PR @mellson and great win for the DX.

apps/extension/client/src/initiateEditor.ts
```diff
@@ -142,13 +138,27 @@ export const initiateEditor = (
...
+      if (document.isDirty) return;
+
       const text = document.getText();
-      const result = parseMachinesFromFile(text);
 
-      if (result.machines.length > 0) {
-        result.machines.forEach((machine, index) => {
+      /*
+       * If we already sent the text, don't send it again.
+       * We need this because onDidChangeTextDocument gets called multiple times on a save.
```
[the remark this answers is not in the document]
farskid: According to the VS Code api docs:
>An event that is emitted when a [text document](https://code.visualstudio.com/api/references/vscode-api#TextDocument) is changed. This usually happens when the [contents](https://code.visualstudio.com/api/references/vscode-api#TextDocument.getText) changes but also when other things like the [dirty](https://code.visualstudio.com/api/references/vscode-api#TextDocument.isDirty)-state changes.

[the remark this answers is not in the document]
farskid: So I guess this is called once when the content is changed and once when the dirty flag changes?

PR #189, opened 2022-07-26
Settings to set the editor theme explicitly

farskid:
Three values are available now in the settings.
Auto means we read the theme kind from the vscode theme and sync that with the editor theme.
Dark means we always render the editor in the dark theme.
Light means we always render the editor in the light theme.
No support for high contract themes yet as the editor doesn't have that concept.

### DIFF
--- a/apps/extension/client/package.json
+++ b/apps/extension/client/package.json
@@ -81,6 +81,16 @@
           "type": "boolean",
           "default": true,
           "description": "Enable nesting the generated typegen files."
+        },
+        "xstate.theme": {
+          "type": "string",
+          "enum": [
+            "auto",
+            "dark",
+            "light"
+          ],
+          "default": "auto",
+          "description": "Color theme for the visual editor. Auto syncs the vscode theme with the editor theme."
         }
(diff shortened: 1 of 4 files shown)


apps/extension/client/package.json
```diff
@@ -81,6 +81,16 @@
...
+        },
+        "xstate.theme": {
+          "type": "string",
+          "enum": [
+            "auto",
+            "dark",
+            "light"
+          ],
+          "default": "auto",
+          "description": "Color theme for the visual editor. Auto syncs the vscode theme with the editor theme."
```
davidkpiano, line 93:
```suggestion
          "description": "Color theme for the visual editor. Auto syncs the VS Code theme with the editor theme."
```

davidkpiano (approved):

apps/extension/client/package.json
mellson: I think this ⬆️ would be a good change, other than that it LGTM 😊

mellson (approved):

PR #190, opened 2022-07-27
Add events leading to final states to the appropriate exit handlers

Andarist:


packages/shared/src/introspectMachine.ts
```diff
@@ -7,6 +7,47 @@ export interface SubState {
...
+  }
+}
+
+function getAllNodesToNodes(nodes: XState.StateNode[]) {
+  const seen = new Set<XState.StateNode>();
+  const result = new Set<XState.StateNode>();
+  for (const node of nodes) {
+    let marker: typeof node | undefined = node;
+    while (marker) {
+      if (seen.has(marker)) {
```
mellson, line 38: I can't stop reading this in my head using Yoda's voice

mellson (approved):
It looks great, you are really good at naming functions, variables etc...

packages/shared/src/introspectMachine.ts
```diff
@@ -7,6 +7,47 @@ export interface SubState {
...
+      );
+      return finalStatesPerRegion.every((perRegion) => !!perRegion.length)
+        ? finalStatesPerRegion.flat()
+        : [];
+    default:
+      return [];
+  }
+}
+
+function getAllNodesToNodes(nodes: XState.StateNode[]) {
```
davidkpiano, line 32: I'm confused, what's the purpose of this function?

Andarist: The name could definitely be improved - i didn't figure out a better one for this. This function grabs all the nodes from a set of nodes to the root.

Screenshot 2022-07-28 at 20 55 56

On this graph, the red circles are the input set and all the green ones are part of the result.

In practice, I don't even need this result more than once. I just end up iterating over it to add the events to all of those states and I could just do it while iterating till the root. I've extracted this to a separate function to split those two concerns here and to make it easier to read but perhaps this wasn't really necessary.

davidkpiano: Ah! This is the `getConfiguration(...)` function in v5.

Andarist: Hm, it's close to that but it's not exactly the same and perhaps my picture was a little bit misleading here. The difference is that a valid configuration always contains states up to the leaves but this accepts any set of nodes as inputs. In practice, I call it with a set of parallel states (so kinda roots of those "leaf parallel" regions).

Andarist: in a machine like this:
```scxml
<scxml>
  <parallel>
    <parallel>
      <state/>
```

I call the inner parallel a "leaf parallel" but its parent parallel state (or any above it) is not considered as such).

Issue #105, opened 2022-03-14
"open visual editor" does nothing


macguffin:
Hi when I click "Open Visual Editor" nothing happens. It did work when I first tried it but now doesn't.
If Iook in Log(Window) I get this error:
```
[renderer3] [error] Cannot get password: Error: Cannot get password
    at I.$getPassword (vscode-file://vscode-/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:1522:45602)
```
Do you think it's a corporate proxy issue? 
Any idea how I get get it to work?

"Open Inspector" is working - assuming that doesn't need to go to the stately.ai site.

Thanks

mattpocock:
Hm, this is likely going to be very difficult to emulate from our end. Can you give us some more information?

Does the webview open? (i.e. does another panel open at all?)
Does your browser open?

mattpocock:
Can you potentially film a video of you attempting to open it?
Is there any output from the extension itself? (in Output/XState)

macguffin:
https://user-images.githubusercontent.com/567652/158996707-272a21fa-c16c-4100-b3af-1a452e8aae2e.mov

It feels to me like a corporate proxy or maybe something to do with the extensions authentication with [stately.ai](http://stately.ai/).

The main bit I want is the viewer which is working fine :).

Here is a video, I’ve just switched to VSC from IDEA (mainly because I like your extension so much). If there is anything else I can provide, let me know.

Also weirdly it did seem to work the very first time I tried it, had a pop up etc and then it was in edit mode. 
Have tried re-installing extension, removing it from the user folder etc.

Is there any way to force re-authentication?

Many thanks

atroshchenkoff:
I have exactly the same problem. The first install worked, but then it stopped working. I tried reinstalling the extension, and even completely removed vscode - nothing helped...

davidkpiano:
Added to Linear: https://linear.app/statelyai/issue/STA-1696/clicking-open-in-visual-editor-not-working-for-some-users

davidkpiano:
Do you see any errors in the developer console? See below for how to open the dev tools in VS Code:

![CleanShot 2022-06-20 at 12 18 25@2x](https://user-images.githubusercontent.com/1093738/174643849-8594706d-683c-43c4-a3bb-2002bc72ebed.png)

atroshchenkoff:
[...]

Issue #102, opened 2022-03-12
Can't sign into Stately in the extension


UberMouse:
Hi,

I can't get the Sign into VSCode link that gets displayed on the stately website when the extension is trying to sign you in, to actually open VSCode. For some reason it just opens Chrome (not my default browser, the vscode extension opens the stately sign in page in Vivaldi) to the new tab page every time I click it.

I'm on Ubuntu 20.04.4 LTS, i3 window manager and the following VSCode version
![image](https://user-images.githubusercontent.com/693020/157995584-9551f893-4bd3-4910-a426-6b57a95e4572.png)

Let me know if there is any other information I can provide, or steps I can perform to help.

mattpocock:
@UberMouse Hm, this is likely going to be tough for us to diagnose. There might be an issue with how Vivaldi interprets vscode's url scheme?

PR #201, opened 2022-08-12
Remove the requirement to authenticate user to use the extension

Andarist:
closes https://github.com/statelyai/xstate-tools/issues/105
closes https://github.com/statelyai/xstate-tools/issues/102


apps/extension/client/testFixture/machine.ts
```diff
@@ -0,0 +1,10 @@
+import { createMachine } from "xstate";
```
Andarist, line 1: it might be good to add this to the repo and then gitignore it so updates to this wont be preserved in git. This is mainly useful (at the moment) as a kitchensink file for testing purposes.

mellson (approved):
Looks great 🎉

Issue #18, opened 2021-10-11
[Bug] Custom guard act as if never used


astahmer:
**Problem:** 
getting a `xxx is never used in the machine definition` when using custom guards (https://xstate.js.org/docs/guides/guards.html#custom-guards)

**Context:**
Using the VS-code extension version `1.2.1` and `xstate` version `4.23.0`.

**Repro:**
When pasting this in the parser (https://xstate-parser-example-site.vercel.app) :

```ts
createMachine(
    {
        initial: "start",
        states: {
            start: {
                on: {
                    MOVE: [
                        { target: "end", cond: { type: "customGuard", param: 123 } },
                        { target: "end", cond: "workingGuard" },
                    ],
                },
            },
            end: {},
        },
    },
    { guards: { customGuard: (ctx) => false, workingGuard: (ctx) => true } }
);
```

We can see that the custom guard is noted as `anonymous`

Silverwolf90:
In the context of typegen, the lack of this support for this leads to type errors. lt would be preferable to even generate `any` or `unknown` until it's properly supported.

<img width="732" alt="image" src="https://user-images.githubusercontent.com/1631625/162047026-5f165d30-ee3d-4285-b2ba-2e0ce6325bc9.png">

CodingDive:
This is a big issue with the typegen currently as any custom guard will not appear in the typegen file. cc @mattpocock 

Also came up in discord https://discord.com/channels/795785288994652170/1001836095253790793

PR #203, opened 2022-08-17
Fixed an issue with parametrized guards not being recognized by typegen

Andarist:
fixes https://github.com/statelyai/xstate-tools/issues/18

### DIFF
--- a/.changeset/few-actors-cover.md
+++ b/.changeset/few-actors-cover.md
@@ -1,0 +1,7 @@
+---
+"@xstate/machine-extractor": patch
+"@xstate/cli": patch
+"stately-vscode": patch
+---
+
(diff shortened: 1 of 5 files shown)


mellson (approved):
Beautifully done!

with-heart:
@Andarist This is literally what I was working on last night 👀 And is a much better solution than what I was working on last night 😆 

Any plans to follow this up with `schema.guards` type safety in `xstate`? Was planning to take a stab at that next if you're not working on it already.

Andarist:
> Any plans to follow this up with schema.guards type safety in xstate? Was planning to take a stab at that next if you're not working on it already.

I would like to avoid introducing this in v4 - right now this requires introducing a new generic type param and such a change affects almost the whole codebase. This is very much planned for v5 though.

PR #202, opened 2022-08-17
improve prettier configs and format files

with-heart:
This PR makes a few changes in order to clean up the repo a bit:

- removed `"prettier": {}` from `package.json` to make sure we're all formatting with `.prettierrc`
- replaced `source.organizeImports` in `.vscode/settings.json` with `prettier-plugin-organize-imports` so all contributors have their imports organized
- added a `pre-commit` hook for `husky`. we already had it installed and `lint-staged` configs exist in `package.json`, we just didn't have a hook to actually do anything with them
- widened `lint-staged` to target `*` files and added the `--ignore-unknown` flag, so now all files recognized by `prettier` will be formatted on commit
- formatted all files with `prettier`

It basically felt like every file I touched ended up with a bunch of unrelated formatting changes, so this should resolve that issue for everyone moving forward.

PR #205, opened 2022-08-17
add pull_request for ci workflow

with-heart:
I noticed that the workflow for #202 ran on my fork but the result wasn't reported in the pull request itself, which makes it difficult for yall to check ci success/failure for outside contributor PRs. I believe this should resolve that and run the workflow on any push or pull request.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,16 @@
 name: CI
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request:
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request && github.event.pull_request.head.sha || env.GITHUB_SHA }}
       - uses: actions/setup-node@v2
         with:
           node-version: 14.x
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,5 @@
 name: CI
-on:
-  push:
-    branches:
-      - "**"
+on: [push, pull_request]
 
 jobs:
   build:


.github/workflows/ci.yml
```diff
@@ -1,8 +1,5 @@
 name: CI
-on:
-  push:
-    branches:
-      - "**"
+on: [push, pull_request]
```
Andarist, line 2: In general, I'm on the fence about how it's best to configure this. With your configuration organization members always create two **identical** runs cause one is created for the push event and the other one is usually created for the pull_request event.

In XState, we've mitigated this by configuring it like this:
```suggestion
on:
  push:
    branches: main
  pull_request
```

It's not quite ideal from my PoV though because I'd really like for any push by organization members to kick off those checks :shrug:

Andarist (commented):
I totally agree with the change. We are using Turborepo at the moment and it requires a secret for its remote caching etc. This won't be available for forks and thus runs might actually fail because of that.

Would you be able to investigate this? Is it possible to workaround this somehow? Could we somehow skip remote caching for those runs without the secret? Is it the only thing that depends on this secret?

If it's not possible to workaround this I would be OK with removing Turborepo from this public repository.

with-heart:
> Would you be able to investigate this? Is it possible to workaround this somehow? Could we somehow skip remote caching for those runs without the secret? Is it the only thing that depends on this secret?

Yeap! I'll take a look and see what's possible

> If it's not possible to workaround this I would be OK with removing Turborepo from this public repository.

If it's not possible, I'll reconfigure things to just use `preconstruct`

.github/workflows/ci.yml
with-heart: Ahhh yeah that makes sense.

with-heart:
From what I can tell, `turbo` will only use remote caching if the `TURBO_TOKEN` and `TURBO_TEAM` env vars are set. It just uses a local cache in `./node_modules/.cache/turbo` otherwise.

PR #217, opened 2022-09-20
Add a telemetry log to track the usage of the VS Code extension

Andarist:


apps/extension/client/src/initiateEditor.ts
```diff
@@ -32,6 +42,20 @@ export const initiateEditor = (context: vscode.ExtensionContext) => {
...
   ) => {
+    if (isTelemetryEnabled) {
+      fetch("https://stately.ai/registry/api/analyze", {
+        method: "POST",
+        body: JSON.stringify([
+          {
+            event: "Opening Editor From VS Code",
+            properties: {
+              time: Math.floor(Date.now() / 1000),
+              distinct_id: machineId,
```
mellson, line 53: Is it possible to get some sort of user id from the Code API? If one user opens multiple machines daily, it would skew our DAU count.

mellson (approved):
I left one comment you can address if you agree; the rest looks good.

apps/extension/client/src/initiateEditor.ts
Andarist: This is inevitable. The same problem happens in a lot of other situations and we can't easily fight it.

mellson (approved):

Issue #200, opened 2022-08-12
VSS extension randomly throw error: Could not find a machine at the current cursor.


nlk-jan:
Hi, 
I am very new to Xstate so apologize for the noob question. 
I cannot get Xstate VSS plugin to open Stately Visualiser reliably.  It randomly appears once or twice (over createMachine), but later it is not showing, and I am getting an error: Could not find a machine at the current cursor.
Thanks
Jan

Andarist:
There is a good chance that the underlying issue is similar to this one https://github.com/statelyai/xstate-tools/issues/194

Issue #194, opened 2022-08-05
Machine ID is truncated when I open the machine in the visual editor


laurakalbag:
https://www.loom.com/share/41c04fbafb964d5091e036dc53afa301

Andarist:
The most likely culprit is the fact that the code lens command embeds the possibly outdated arguments:
https://github.com/statelyai/xstate-tools/blob/53ba81ba15f991952f220c0b519225328f06b9f4/apps/extension/server/src/server.ts#L164-L173

The fix would be to limit the arguments and grab the config "on demand", when the user actually clicks on the code lens. This way we'd guarantee that we grab the most up-to-date config etc

Andarist:
We can't be completely sure as we never confirmed what has causes this issue but I strongly believe that there is non-zero chance that this is being fixed by: https://github.com/statelyai/xstate-tools/pull/220

Issue #42, opened 2022-01-29
Throttle configuration for typegen autosave 


slowselfip:
With the new typegen feature the extension is doing some "throttled auto save as you type". This can get a bit heavy on the CPU if there's for example a dev-server running in the background watching files and doing recompiles.

Not exactly sure what the best solution would be, but if it would be possible to make the auto save feature configurable that could help.

A workaround is to close any file watchers whilst editing the machine config.

And I must say that the typegen support is probably the biggest leap in developer experience I've ever encountered! Really impressed with what you've achieved. Great work!

mattpocock:
Thank you @slowselfip! My plans are to add a configuration option to VSCode that would let you choose between:

Typegen running on throttle (allows for very quick TS updates, but CPU intensive)
Typegen running on save (TS updates when you save the file, similar to if you were running a CLI)

with-heart:
> Typegen running on save (TS updates when you save the file, similar to if you were running a CLI)

This would be perfect! Having `jest --watch` re-run the tests every time I type a few letters is not ideal 😁

raghavraman:
👋  Any updates on this. Would greatly benefit from the save the file option. Thanks

Andarist:
I believe that this is being fixed by https://github.com/statelyai/xstate-tools/pull/220 because with the changes in that PR we only update the typegen files when their output changes.

PR #220, opened 2022-09-21
Move source parsing to server

Andarist:
closes https://github.com/statelyai/xstate-tools/issues/42
closes https://github.com/statelyai/xstate-tools/issues/194
closes https://github.com/statelyai/xstate-tools/issues/200
closes https://github.com/statelyai/xstate-tools/issues/230


packages/shared/src/__tests__/__snapshots__/getTypegenOutput.test.ts.snap
```diff
@@ -155,21 +155,22 @@ exports[`getTypegenOutput entry-action-on-ancestor-of-targeted-descendant 1`] =
 export interface Typegen0 {
   "@@xstate/typegen": true;
   internalEvents: {
+    GO_STRING: { type: "GO_STRING" };
```
Andarist, line 158: this change is worrying, I need to investigate this

apps/cli/src/__tests__/__examples__/action-backtick.ts
```diff
@@ -1,13 +1,13 @@
-import { createMachine } from "xstate";
+import { createMachine } from 'xstate';
```
with-heart, line 1: If you ever reach a point where you don't have much work in flight, we should update and merge #202 so we don't have this formatting crap littering so many of your PRs 😆

Andarist: Rebased and fixed :P

PR #221, opened 2022-09-21
create `eslint-plugin-xstate` package

### LINKED ISSUE #216: create `eslint` config for v5
I've been thinking about how cool it would be to provide tooling to assist with (eventual) user migration to v5. One of the things that could be quite helpful would be an `eslint` config containing rules that help the user work through all of the breaking changes or utilize new functionality.

Here's what I think those rules might look like, using the [`xstate@5.0.0-alpha.0` release notes](https://github.com/statelyai/xstate/releases/tag/xstate%405.0.0-alpha.0) as a starting point.

| rule | description |
|------|-------------|
| `avoid-context-spread` | Since `machine.withContext` now permits partial context, no need to spread `machine.context` |
| `no-use-atomic-internal` | `internal` property no longer has an effect on atomic state nodes |
| `no-use-cond` | `cond` has been renamed to `guard` |
| `no-use-deprecated-config-properties` | `onEntry`, `onExit`, `parallel`, and `forward` have been removed |
| `no-use-eventless-on-transition` | Eventless transitions must now be specified using `always` instead of `on['']`. |
| `no-use-emitted-from` | `EmittedFrom` has been renamed to `SnapshotFrom` |
| `no-use-factory-context-arg` | The 3rd arg (`context`) has been removed from and `createMachine()` |
| `no-use-generic-state-schema` | `StateSchema` has been removed from all generics |
| `no-use-guard-in` | The `in` property for transitions has been removed and replaced with guards. Prefer `stateIn()` or `not(StateIn())`
| `no-use-machine` | `Machine` has been removed; use `createMachine` instead |
| `no-use-machine-transition-context-arg` | 3rd arg (`context`) to `machine.transition` has been removed. Can use `State.from('state', {})` as 1st argument instead. |
| `no-use-service-batch` | `service.batch(events)` has been removed |
| `no-use-service-children` | children can only be accessed from `state.children` |
| `no-use-service-execute` | `service.execute` has been removed |
| `no-use-service-onChange` | `service.onChange` has been removed in favor of `service.onTransition` or `service.subscribe` |
| `no-use-service-onEvent` | `service.onEvent` has been removed in favor of `service.onTransition` or `service.subscribe` |
| `no-use-service-onSend` | `service.onSend` has been removed in favor of `service.onTransition` or `service.subscribe` |
| `no-use-service-send-type-payload` | `service.send(type, payload)` is no longer supported |
| `no-use-spawn-import` | `spawn` is now available in 3rd arg to `assign` |
| `no-use-state-activities` | `state.activites` has been removed |
| `no-use-state-children-direct-reference` | `state.children` is now a mapping of invoked actor IDs to their `ActorRef` and should never be referenced directly. _Not really sure about the name or what this rule would even do, but wanted to make a note of it_ |
| `no-use-state-events` | `state.events` has been removed |
| `no-use-state-history` | `state.history` has been removed |
| `no-use-state-historyValue` | `state.historyValue` is considered internal |
| `no-use-state-node-isTransient` | `stateNode.isTransient` has been removed |
| `no-use-state-node-version` | `stateNode.version` has been removed. `version` is only available on root `machine` node |
| `prefer-spawn-implementation-name` | Prefer referencing actors defined in `config.services` by name (`spawn('promiseActor')` instead of `spawn(promiseActor)`) |
| `prefer-wildcard-event-descriptors` | Detects multiple namespaced actions with the same config object that could be combined (?) |
| `rename-machine-withConfig-to-provide` | `machine.withConfig(…)` -> `machine.provide(…)` |
| `require-compound-state-initial-key` | Compound states now require an `initial` key (did they not already?) |
| `require-object-context` | `machine.context` is now required to be an `object` |
| `require-parameterized-actions-params` | Parameterized actions now require a `params` property (`{ message: 'Hello' }` -> `{ params: { message: 'Hello' }}`) |
| `require-parameterized-guards-params` | Guard parameters should now be placed in object at `params` key (`{ minQueryLength: 3 }` -> `{ params: { minQueryLength: 3 }}`) |

with-heart:
Hi all! This draft PR contains the start of the `eslint-plugin-xstate` package and will eventually resolve #216.

Currently this PR contains implementations for a few rules:

- `avoid-context-spread`
- `no-cond`
- `no-emitted-from`
- `no-factory-context-arg`
- `require-object-context`
- `require-parameterized-actions-params`
- `require-parameterized-guards-params`

Having briefly talked to @Andarist about it, I'm not entirely sure an `eslint` plugin/config makes sense for his vision of XState tooling. It's possible that these rules (and others listed in the issue) would be more useful if implemented as part of the VSCode extension rather than in `eslint`.

At the very least, this has been a fun exploration of custom `eslint` rules for me and might give us a sense of the type of logic we'd need to implement similar functionality in the extension.


package.json
```diff
@@ -87,10 +87,10 @@
     "@preconstruct/cli": "^2.1.0",
     "@types/node": "^16.0.1",
     "@typescript-eslint/eslint-plugin": "^4.23.0",
-    "@typescript-eslint/parser": "^4.23.0",
+    "@typescript-eslint/parser": "^5.37.0",
     "concurrently": "6.2.0",
     "esbuild": "^0.14.48",
-    "eslint": "^7.26.0",
+    "eslint": "^8.0.0",
```
with-heart, line 93: We have the config in `.eslintrc.js` commented out so updating these doesn't affect anything

packages/eslint-plugin-xstate/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "peerDependencies": {
+    "@typescript-eslint/parser": "^5.0.0",
+    "eslint": "^6.0.0 || ^7.0.0 || ^8.0.0",
+    "typescript": "*"
+  },
+  "peerDependenciesMeta": {
+    "typescript": {
+      "optional": true
+    }
+  }
```
with-heart, line 31: I'm not entirely sure these dependencies are correct if we end up publishing this package. I somewhat copied the dependencies setup from [`@typescript-eslint/eslint-plugin`](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin), but they also don't have any of these as `devDependencies` which `yarn` whined about 🤷

packages/eslint-plugin-xstate/src/rules/avoid-context-spread.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import rule from "./avoid-context-spread";
+
+const ruleTester = new ESLintUtils.RuleTester({
+  parser: "@typescript-eslint/parser",
+});
+
+ruleTester.run("avoid-context-spread", rule, {
+  valid: [
+    {
+      code: outdent`
```
with-heart, line 12: Using `outdent` allows the code blocks to be displayed correctly in test output. Not sure if there's a better way to handle that but without it, the indentation was a mess.

with-heart:
Would love to have a discussion on `eslint` vs. other ways to implement.

My ultimate goal in playing with this stuff was to use it as the basis for a v5 codemod that automatically migrates codebases (as much as possible, anyway). Perhaps it'd be better to just start with that, or figure out ways to implement this type of stuff through the extension instead of `eslint`?

davidkpiano:
> Would love to have a discussion on `eslint` vs. other ways to implement.
> 
> My ultimate goal in playing with this stuff was to use it as the basis for a v5 codemod that automatically migrates codebases (as much as possible, anyway). Perhaps it'd be better to just start with that, or figure out ways to implement this type of stuff through the extension instead of `eslint`?

I would say `eslint` is preferable.

EightArmCode:
Someone already wrote an `eslint-plugin-xstate` [FWIW](https://github.com/rlaffers/eslint-plugin-xstate)

PR #222, opened 2022-09-22
Parse TemplateLiteral actions

### LINKED ISSUE #213: Typegen does not work with template literals in actions
Screenshot 2022-09-14 at 07 58 10

with-heart:
This PR closes #213 by updating the `TemplateLiteral` parser to use `path.evaluate()` to retrieve its string representation and creating a new `ActionAsTemplateLiteral` parser.

Note that this PR is branched from #214 as we need access to `path`.

See 8bcdb3d45a53e1385dd05826dd2a1798291146be and 9d32935 for relevant changes

### DIFF
--- a/packages/machine-extractor/src/scalars.ts
+++ b/packages/machine-extractor/src/scalars.ts
@@ -21,7 +21,7 @@ export const StringLiteral = unionType([
     maybeIdentifierTo(
       createParser({
         babelMatcher: t.isStringLiteral,
-        parseNode: (path): StringLiteralNode => {
+        extract: (path): StringLiteralNode => {
           return {
             value: path.node.value,
             node: path.node,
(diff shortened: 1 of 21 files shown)


packages/machine-extractor/src/scalars.ts
```diff
@@ -4,87 +4,88 @@ import {
...
 });
 
 export const TemplateLiteral = maybeTsAsExpression(
   maybeIdentifierTo(
     createParser({
       babelMatcher: t.isTemplateLiteral,
-      parseNode: (node) => {
-        let value = "";
+      extract: (path) => {
+        const evaluated = path.evaluate();
```
with-heart, line 81: We might need to do some additional work here since `evaluate` might not work correctly if it encounters unidentifiable `quasis`. Maybe we just return `null` if that's the case because it's likely an issue on the user's end?

PR #225, opened 2022-09-29
Capture any string in the meta field of a machine or state

abrgr:
Currently, only description is captured from meta. It would be super helpful to capture any string meta field.


Andarist:
What does this solve? Where do you expect to see those extra keys?

abrgr:
At the moment, I would like to use them [here](https://github.com/simplystated/xstate-to-gherkin) to generate gherkin tests from an xstate statechart. Meta keys are generally useful for any type of statechart reflection like this. If they are captured by the machine-extractor, it becomes much easier to build these types of tools that rely on state metadata.

davidkpiano:
This is a favorable change, since `meta.description` should be deprecated anyway in favor of `description`.

Andarist:
I'm just slightly worried about the ripple effect of this change at the moment. I'm unsure how the previous thing was exactly utilized by our system. The current API of this package is really coupled at times to the underlying AST - and in fact, this change is a breaking change for this package. 

I don't mind this change but I would have to do a deeper analysis to see if this would be a breaking change *for us*.

abrgr:
Totally makes sense. Let me know if I can be helpful in analyzing that.

If it helps, the previous behavior was to capture `meta.description` but to ignore any other field on `meta`. So if a file had a machine that defined

```
meta: {
  abc: "abc"
}
```

`parseMachinesFromFile` would return a machine/state without a meta object.

If a file had a machine that defined:

```
meta: {
  description: "my description",
  abc: "abc"
}
```

`parseMachinesFromFile` would return a machine/state with meta of only `{ description: "my description" }`.

Andarist:
The breaking change I had in mind is about the shape of the extracted node - unfortunately, the AST shape leaks there. Now you have `.properties` property there, whereas previously it was `.description`.

abrgr:
@Andarist - 2 thoughts:

1. I can make StateMeta in packages/machine-extractor/src/meta.ts a proxy that throws on any access to .description and run any tests or anything you had in mind to see if that's accessed.
2. I can modify StateMeta in packages/machine-extractor/src/meta.ts to return an AST with both `.properties` and `.description` if `.description` exists.

Would either of these be helpful?

Issue #228, opened 2022-10-14
prettier-plugin-svelte prevents the VS Code extension to write typegen files


phcoliveira:
Hello there!

With a brand new SvelteKit installation, the XState Visual Code extension does not write typegen files out of the box.

As pointed out by @mellson, the extension can resume working when the Prettier plugin `prettier-plugin-svelte` is removed and VS Code is restarted.

Everything else regarding the extension still works:

1. After saving the machine file, the following is added `tsTypes: {} as import("./foo-machine.typegen").Typegen0`
2. The functionalities `Open Visual Editor` and `Open Inspector` 
3. A comment is added: `/** @xstate-layout N4Igp.... */ 

I am not affirming that the bug is to be fixed by the extension. Perhaps there is something wrong with the Prettier plugin.

Here is a reproducible repo:
https://github.com/phcoliveira/sveltekit-xstate

mellson:
Thank you @phcoliveira! - [Discord thread for more context here](https://discord.com/channels/795785288994652170/1030398468654628864).

Andarist:
This will get fixed next week since we are removing the dependency on Prettier from the VS Code extension.

phcoliveira:
Same background you two? Doesn't seem like a coincidence.

Andarist:
We are both members of The Brick Gang 😉

Andarist:
We are removing the dependency on Prettier in [this PR](https://github.com/statelyai/xstate-tools/pull/229) so I believe that your issue is going to be fixed soon. Sorry for the delay!

Issue #223, opened 2022-09-22
File is not updated when you close the tab


lokhmakov:
When you close the tab, the file doesn't accept the changes. It is possible to lose changes in this way.

If you open the file after editing and click on the visualiser, the update will happen.

Other issues noticed:

1. You can't select all text in input fields in the renderer by CMD + A
2. If you edit in the file, the changes are erased.

![bug-2022-09-22](https://user-images.githubusercontent.com/94939/191810465-a8cf4530-e2fc-4574-bba7-ed0b349bfb3c.gif)

Andarist:
Thank you for your report. I've found the VS Code APIs that we can use to fix this.

Andarist:
I believe that this is being fixed by https://github.com/statelyai/xstate-tools/pull/229 since I've switched there to directly affect the given URI through a `WorkspaceEdit`: https://github.com/statelyai/xstate-tools/pull/229/files#diff-7fa1e063efd6c4eb2e3c9524b5ba5528fa926c78cdfa638c759312ebcffc40beR320-R321 . Previously we were first looking for an active text editor for a given URI ([here](https://github.com/statelyai/xstate-tools/pull/229/files#diff-470fd101b7177066618766965e53c300eee26d04ddc0fdc97eeea23dc7524bc8L22-L24)) but that was no longer available when you closed your text editor containing the displayed machine.

Andarist:
Actually, #229 only has fixed part of the problem. The other part is being fixed by https://github.com/statelyai/xstate-tools/pull/256

Issue #211, opened 2022-09-09
Editing a machine visually can result in information being lost in the code


jamesbvaughan:
I have a machine that I edit both directly in the code and in the visual editor in the VS Code extension.

After editing the machine via the visual editor, the following are removed from the machine source:
- comments
- state metadata (I see that this is tracked in https://github.com/statelyai/xstate-tools/issues/104)
- [forbidden transitions](https://github.com/statelyai/xstate-tools/issues/104)

Here is a simple machine that demonstrates these issues:
```typescript
import { createMachine } from "xstate";

createMachine({
  id: "example",
  initial: "first",
  states: {
    first: {
      meta: {
        example: "example",
      },
      on: {
        next: {
          target: "second",
        },
      },
    },
    // This is a comment
    second: {
      on: {
        next: {
          target: "third",
        },
        backToFirst: undefined,
      },
    },
    third: {},
  },
  on: {
    backToFirst: {
      target: ".first",
    },
  },
});
```

Here is a diff showing the changes made to this machine by simply opening up the visual editor (and running `prettier`):
```diff
 import { createMachine } from "xstate";

+/** @xstate-layout N4IgpgJg5mDOIC5RgB4EMC2AHANmAdAGYCWATrAC4DEAdqhYqFgPazEXHM2MgqIBMANgCM+YQFZBQgCwBOAMzz+wgAzyA7ABoQAT0Szx+cSpPjh82QA4VQxQF8721JlwFYYAMZcItejxZsHFw8fAjS4vz4guqy6hHqlgqW0pbaeggGRiYqZhbWtvIOTujYeFQARmgeANYAKswAYmSU-qzsnNxIvIjS-Gn6hsam5lY2gvZFIDTMEHA8zqUEJOQMXQHtwV2hwurqUZJCKuqC4vmpuvoq+NkqgqPSR1Kykwuu+O5eNBCtgR0hiPJpIIjAlhCJhDtxEo1P0MlcbrYFKdVJYXiU3hQABZkb5rNpBTqgUJQyxiFRglKxDS3cSw2Tw7KI+TIlSoxzgdF4H4bQndBDCc7pQQOBxAA */
 createMachine({
   id: "example",
   initial: "first",
   states: {
     first: {
-      meta: {
-        example: "example",
-      },
       on: {
         next: {
           target: "second",
         },
       },
     },
-    // This is a comment
     second: {
       on: {
         next: {
           target: "third",
         },
-        backToFirst: undefined,
       },
[...]

Issue #163, opened 2022-06-14
Optional automatic formatting with prettier


Silverwolf90:
When you open a machine with the visual editor, it automatically formats the machine using prettier and there is no way to opt out of this behavior. I don't use prettier so it would be better to make this optional.

Andarist:
What kind of formatting would you like to have by default? Are you using any other formatter?

mattpocock:
The way we have it set up now, we _need_ to use some kind of default formatting - otherwise it'll just present as `JSON.stringify(config, null, 2)`.

I.e. any time you have generated code, the thing that generates the code needs to make choices about how that code displays.

Andarist:
There are various ways to achieve the result though. A simple formatter could just work similarly to `JSON.stringify` (we probably can't use it directly because we might need to handle things that are not fully serializable, especially in the future) with indentation detected based on the file and the position in the file.

Silverwolf90:
Yeah I don't use an auto-formatter at all (except for eslint which is pretty limited), but I see the problems here.

Perhaps what feels most problematic is that you format the machine before I've made any material changes via the visual editor -- so even if I just want alter the layout in the editor, I incur formatting changes that I don't want.

Moving from there, it seems that _ideally_  the only parts that should be formatted are parts that have been introduced via the visual editor, and avoid formatting changes to parts that haven't. I realize that would be pretty difficult 😅

I guess these kind of problems are what building this bidirectional code<->visual editing  pretty difficult (I remember Dreamweaver pretty fondly though 😂)

Andarist:
> Moving from there, it seems that ideally the only parts that should be formatted are parts that have been introduced via the visual editor, and avoid formatting parts that haven't. I realize the complexity of that makes it very difficult to build 😅

Ye - this isn't impossible but definitely more complex than the current solution. Given our small team right now and the relatively low priority of this, I wouldn't expect this to happen any time soon. A good thing though is that this is an open-source repository - so... sky's the limit 😉

Andarist:
The dependency on Prettier is being removed (for the most part) in this PR: https://github.com/statelyai/xstate-tools/pull/229

PR #229, opened 2022-10-14
Integrate machine edits with the Studio, apply received edits in the VS Code

Andarist:
closes https://github.com/statelyai/xstate-tools/issues/163
closes https://github.com/statelyai/xstate-tools/issues/66
closes https://github.com/statelyai/xstate-tools/issues/104
closes https://github.com/statelyai/xstate-tools/issues/134
closes https://github.com/statelyai/xstate-tools/issues/211
closes https://github.com/statelyai/xstate-tools/issues/223
closes https://github.com/statelyai/xstate-tools/issues/228
closes https://github.com/statelyai/xstate-tools/issues/238
closes https://github.com/statelyai/xstate-tools/issues/239


apps/extension/server/src/server.ts
```diff
@@ -565,12 +559,39 @@ connection.onRequest('getMachineAtCursorPosition', ({ uri, position }) => {
   };
 });
 
+connection.onRequest('applyMachineEdits', ({ machineEdits }) => {
+  if (!displayedMachine) {
+    throw new Error(
+      '`applyMachineEdits` can only be requested when there is a displayed machine',
```
davidkpiano, line 565: Where does this error get displayed?

apps/extension/server/src/server.ts
```diff
@@ -565,12 +559,39 @@ connection.onRequest('getMachineAtCursorPosition', ({ uri, position }) => {
...
+    );
+  }
+
+  const modified = documentsCache
+    .get(displayedMachine.uri)!
```
davidkpiano, line 570: Nit: can we guarantee this always exists?

apps/extension/server/src/server.ts
Andarist: Excellent question - it won't be displayed anywhere. This error has just been added to satisfy TS as it should be almost impossible to end up with it. I don't like "almost" in this sentence though and gonna try to look into providing stronger guarantees around this.

apps/extension/server/src/server.ts
Andarist: Yes, we can't have a displayed machine without having a cached document for it.

mellson (approved):

PR #233, opened 2022-10-19
Add description updates and state type changes

Andarist:


packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -12,6 +12,10 @@ import { toMachineConfig, ToMachineConfigOptions } from './toMachineConfig';
 import { TransitionConfigNode } from './transitions';
 import { Comment } from './types';
 
+function last<T extends ReadonlyArray<any>>(
```
davidkpiano, line 15: Fancy!

mellson (approved):

PR #236, opened 2022-10-26
Fixed an issue with renaming states with empty keys

Andarist:


packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -800,99 +800,67 @@ export class MachineExtractResult {
             ? stateProp.key.value
             : null;
 
+          if (oldName === null) {
+            throw new Error('Could not find the old name of the state');
+          }
```
Andarist, line 805: this just moved this error to act as an early~ return and it now compared `oldName` against `null`, instead of checking its truthiness

davidkpiano (approved):
Test LGTM; can't speak too much to the code

PR #242, opened 2022-11-10
Send potential errors to the Studio

mellson:
This PR adds a banner to the visual editor where we show any potential error messages coming from parsing machines.

**Example**

https://user-images.githubusercontent.com/167574/201145169-840b0a83-4382-4090-8e39-d8665deadff0.mp4


apps/extension/client/src/initiateEditor.ts
```diff
@@ -270,7 +282,7 @@ const machine = createMachine(
           ),
       onDisplayedMachineUpdatedListener:
```
Andarist, line 283: this should be renamed to `onServerNotificationListener` or something

apps/extension/client/src/initiateEditor.ts
```diff
@@ -270,7 +282,7 @@ const machine = createMachine(
...
         ({ extensionContext, languageClient }) =>
-        (sendBack) =>
+        (sendBack) => {
           registerDisposable(
```
Andarist, line 286: the return values from both of those `registerDisposable` calls here should be returned. This would look smth like this:
```ts
const dispose1 = registerDisposable(...)
const dispose2 = registerDisposable(...)
return () => {
  dispose1()
  dispose2()
}
```

packages/machine-extractor/src/utils.ts
```diff
@@ -431,3 +431,13 @@ function hash(str: string): string {
     .toString(32)
     .substring(1, 10);
 }
+
+export const isErrorWithMessage = (
```
Andarist, line 435: I'd move this to be local util, close to the server stuff (as it's used there). We could have this in some kind of a shared thing but I don't think that machine-extractor package is that. This util falls outside of its domain and shouldn't be part of its public API

apps/extension/server/src/server.ts
```diff
@@ -259,6 +265,11 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
       });
     }
   } catch (e) {
+    if (isErrorWithMessage(e)) {
+      connection.sendNotification('extensionError', {
+        message: e.message,
+      });
+    }
```
Andarist, line 272: If we don't trust the thrown thing here to be an error then we should also handle the other case. Something like this would do the trick:

```suggestion
    connection.sendNotification('extensionError', {
      message: isErrorWithMessage(e) ? e.message : 'Unknown error',
    });
```

packages/shared/src/types.ts
```diff
@@ -115,4 +115,7 @@ export interface NotificationMap {
     uri: string;
     types: TypegenData[];
   };
+  extensionError: {
```
Andarist, line 118: I'd probably rename this to be a little more specific - like maybe `extractionError` or something like that

apps/extension/server/src/server.ts
```diff
@@ -216,6 +217,11 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
 
     if (displayedMachine?.uri === textDocument.uri) {
       const machineResult = machineResults[displayedMachine.machineIndex]!;
+      // If we got this far we can safely assume that the machine config is valid and we can clear any potential errors
+      connection.sendNotification('extensionError', {
+        message: undefined,
+      });
```
Andarist, line 223: This is true and it works fine but I feel like when debugging what's going on in the server-client communication this might get a little bit spammy. I'd try to only send this out if we actually have a displayed error in the first place.

apps/extension/server/src/server.ts
```diff
@@ -259,6 +265,11 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
       });
     }
   } catch (e) {
```
Andarist, line 267: This currently works differently for config-related errors (eg. when a transition targets a state that doesn't exist) when we have `tsTypes` and when we don't have them.

With `tsTypes` it crashes because `getTypegenData` call leads to `createIntrospectableMachine` and it crashes there or in some code that introspects that result. In such a case we nicely display the error in the Editor (with a caveat that you can introduce a config-related error to a different machine in a file and you'd still see the error in the Editor when displaying something else, this could have been fixed).

Without `tsTypes` the invalid config gets sent to the Editor and it crashes there in the `UPDATE_CONFIG` handler within the `fromMachineConfig` call. No error is displayed in this case.

It could be easier to patch this in the Editor but I feel like we should really fix this here. There is currently a no good way to check for those kinds of errors though. You could try to abuse the `createIntrospectableMachine` to check for this though.

What you'd have to do is to map those results differently:
https://github.com/statelyai/xstate-tools/blob/83d5b671b67f378366939312c3ca253510c6fb1b/apps/extension/server/src/server.ts#L199-L210

We need to end up with an array like this `Array<{ types?: TypegenData, configError?: string }>` and handle that appropriately in logic that follows this `.map`.

apps/extension/client/src/initiateEditor.ts
```diff
@@ -284,7 +296,23 @@ const machine = createMachine(
...
+
+          const extensionErrorDisposable = registerDisposable(
+            extensionContext,
+            languageClient.onNotification('extractionError', ({ message }) => {
+              sendBack({
+                type: 'EXTENSION_ERROR',
+                message,
+              });
+            }),
+          );
```
Andarist, line 309: After renaming the notification this disposable should match the notification's name - so it should be `extractionErrorDisposable`. I would probably also rename the `EXTENSION_ERROR` to `EXTRACTION_ERROR`

apps/extension/server/src/server.ts
```diff
@@ -195,17 +199,30 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
...
+
+    // Iterate non-typed machines and throw if there are any errors.
+    machineResults
+      .filter(
+        (machineResult): machineResult is NonNullable<typeof machineResult> =>
+          machineResult?.machineCallResult.definition?.tsTypes?.node ===
+          undefined,
+      )
+      .forEach((machineResult) =>
+        introspectMachine(createIntrospectableMachine(machineResult) as any),
```
Andarist, line 225:
```suggestion
        // for the time being we are piggy-backing on the fact that `createMachine` called in `instrospectMachine` might throw for invalid configs
        introspectMachine(createIntrospectableMachine(machineResult) as any),
```

apps/extension/server/src/server.ts
```diff
@@ -34,9 +36,11 @@ import { getCursorHoverType } from './getCursorHoverType';
 import { getDiagnostics } from './getDiagnostics';
 import { getReferences } from './getReferences';
 import { CachedDocument } from './types';
+import { isErrorWithMessage } from './utils';
 
 let hasConfigurationCapability = false;
 let hasWorkspaceFolderCapability = false;
+let hasActiveError = false;
```
Andarist, line 43: Because this variable is "global" right now it can potentially lead to bugs down the road. I think that right now it won't necessarily lead to a bug but there is a data inconsistency problem here and, in the future, someone could accidentally use this incorrectly and thus introduce an actual bug without knowing it.

The problem is that the user can switch to displaying a different machine through a code lens. This requests the machine's data from the server and forwards that to the editor using `UPDATE_CONFIG` event. Based on that we already clear the error in the editor so it won't be displayed there but the `hasActiveError` variable here won't be cleaned up. Note that such a scenario isn't related to the document change and thus `handleDocumentChange` won't be called and won't reset the `hasActiveError` to `false` just yet.

For that reason, I would track those errors more locally - per machine - in the `documentsCache`. Syntax errors have to be added to all machines in the file but config-level errors should only be added to a single machine.

Andarist: And actually, there is a bug here now - because config-level errors are global now, we might introduce a config-related error to a machine that is not displayed and yet we'd display the error there. This should be avoided.

We can't as easily rely on a catch-all `catch` for those config-related errors. We should actually catch them per `introspectMachine` call while mapping this. Since `getTypegenData` also calls that `introspectMachine` under the hood - it would probably be best to map machines in a loop to an array close to what I've proposed [here](https://github.com/statelyai/xstate-tools/pull/242#discussion_r1020237668):
> We need to end up with an array like this Array<{ types?: TypegenData, configError?: string }> and handle that appropriately in logic that follows this .map.

apps/extension/server/src/server.ts
Andarist: I wonder. Do we actually need to call `introspectMachine` here? Or is it enough to call `createInstrospectableMachine`? I'm not sure which one actually throws those errors.

mellson: In my testing, I found that we do need to call `introspectMachine` before it started throwing on errors

apps/extension/server/src/server.ts
mellson: Thanks for the input; it makes sense to handle this more locally. I've tried to do a version in the latest commit; let me know what you think.

packages/machine-extractor/src/types.ts
```diff
@@ -32,7 +32,7 @@ export interface Comment {
 }
 
 export interface FileExtractResult {
-  machines: (MachineExtractResult | undefined)[];
+  machines: MachineExtractResult[];
```
Andarist, line 35: I don't think this is a correct change. This `FileExtractResult["machines"]` type is also used [here](https://github.com/statelyai/xstate-tools/blob/f99f685e1f9d85749c7813e4b26133b6c2b2c4d0/packages/machine-extractor/src/extractMachinesFromFile.ts#L18) and I don't think that `MachineExtractResult` is guaranteed there. I actually don't like this and it would be cool to refactor this in the future but that would be a bigger refactor

apps/extension/client/src/initiateEditor.ts
```diff
@@ -127,12 +127,21 @@ type DisplayedMachineUpdated = {
   implementations: ImplementationsMetadata;
 };
 
+type ExtensionError = {
```
Andarist, line 130:
```suggestion
type ExtractionError = {
```

apps/extension/server/src/server.ts
```diff
@@ -192,21 +200,42 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
     }
     const [settings, machineResults] = await Promise.all([
       getDocumentSettings(textDocument.uri),
-      filterOutIgnoredMachines(extracted).machines,
+      filterOutIgnoredMachines(extracted).machines.filter(isMachineResult),
```
Andarist, line 203: `filterOutIgnoredMachines` is already filtering this:
https://github.com/statelyai/xstate-tools/blob/a846c768072524f86a51a20b379480e932f768fb/packages/shared/src/filterOutIgnoredMachines.ts#L15-L17 
```suggestion
      filterOutIgnoredMachines(extracted).machines,
```

mellson: True, but it's defined as possibly being undefined now with the revert of the type of machine in the `FileExtractResult` interface. So maybe I should make `filterOutIgnoredMachines` return a new type where the machines can't be undefined?

apps/extension/server/src/server.ts
```diff
@@ -192,21 +200,42 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
...
-          !!machineResult?.machineCallResult.definition?.tsTypes?.node,
-      )
-      .map((machineResult, index) =>
-        getTypegenData(
-          UriUtils.basename(URI.parse(textDocument.uri)),
-          index,
-          machineResult,
-        ),
-      );
+    // Iterate over the machine results and add possible typegen data plus config errors
```
Andarist, line 206: IMHO this repeats the implementation too much and might have consistency issues with evolving implementation. I'd prefer avoiding comments that don't add additional information to the code or ones that explain particularly complex logic etc.

apps/extension/server/src/server.ts
mellson: I've tried to create a generically typed version of `FileExtractResult` in the latest commits, maybe that's better?

apps/extension/server/src/server.ts
```diff
@@ -192,21 +200,42 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
...
+    const machinesWithPossibleTypesAndErrors = machineResults.map(
```
Andarist, line 206: Maybe something like `extractionResults`?
```suggestion
    const extractionResults = machineResults.map(
```

packages/machine-extractor/src/types.ts
```diff
@@ -31,8 +30,8 @@ export interface Comment {
   type: 'xstate-ignore-next-line' | 'xstate-layout';
 }
 
-export interface FileExtractResult {
-  machines: (MachineExtractResult | undefined)[];
+export interface FileExtractResult<T> {
```
Andarist, line 33:
```suggestion
export interface FileExtractResult<T extends MachineExtractResult | undefined = MachineExtractResult | undefined> {
```

packages/machine-extractor/src/extractMachinesFromFile.ts
```diff
@@ -1,11 +1,12 @@
+import { MachineExtractResult } from '.';
```
Andarist: this creates a cyclic dependency (`src/index` depends already on this file, `src/extractMachinesFromFile`, and now this file depends on `src/index` too). I would reach to the actual source for this import:

 ```suggestion
import { MachineExtractResult } from './MachineExtractResult';
```

Comment (mellson) [reply packages/machine-extractor/src/extractMachinesFromFile.ts]: I removed the type in the latest commit as you added the default, and thanks for that 👍🏻

File: apps/extension/server/src/server.ts [hunk 18]
```diff
@@ -215,23 +238,43 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
...
+
+      if (configError) {
+        connection.sendNotification('extractionError', {
+          message: configError,
         });
+      } else {
+        // If we got this far we can safely assume that the machine config is valid and we can clear any potential errors
+        connection.sendNotification('extractionError', {
+          message: undefined,
+        });
```
Andarist, line 252: this again got a little bit "chatty"/noisy. You had a good filtering in place here:
https://github.com/statelyai/xstate-tools/blob/cd9dfcb579b20649e5eb069814df4d7f9ad0c3a0/apps/extension/server/src/server.ts#L238-L243

but that got lost after making the errors more local. I propose to keep `extractResults` in the `CachedDocument` to always get all the information that we need about the previous document. This would replace `CachedDocument['machineResults']` as those machine results are already part of the `extractResults`

mellson: Good catch 🎣 I've just committed a less chatty version

apps/extension/server/src/getCursorHoverType.ts
```diff
@@ -61,7 +62,9 @@ export const getCursorHoverType = (
       machine: MachineExtractResult;
     }
   | void => {
-  for (const machineResult of machineResults) {
+  for (const machineResult of extractionResults.map(
+    (extractResult) => extractResult.machineResult,
+  )) {
```
Andarist, line 67:
```suggestion
  for (const { machineResult } of extractionResults) {
```

Andarist (approved):

PR #251, opened 2022-11-18
Save syntax errors on previous extraction result

mellson:
This PR adds any syntax errors to the previous extraction result. Which will make it possible to clear the error again.


apps/extension/server/src/server.ts
```diff
@@ -308,9 +308,17 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
       });
     }
   } catch (e) {
-    if (displayedMachine?.uri === textDocument.uri) {
+    if (displayedMachine?.uri === textDocument.uri && isErrorWithMessage(e)) {
```
Andarist, line 311: A syntax error is relevant for all machines in the cached document - so to keep data consistent I would loop through all current extraction results and add this error to all of them

Andarist: The error notification itself is only relevant to the displayed machine though (or at least to its document, cause we could introduce a syntax error into a document that we don't currently display)

mellson: Excellent point; I've tried to solve this by adding the syntax error on the top level in the CachedDocument

Andarist: Two small issues here:
- syntax error might still be somehow relevant for the document regardless of `displayedMachine?.uri === textDocument.uri`. I would assign that to the document at all times **but** only dispatch the `extractionError` notification when `displayedMachine?.uri === textDocument.uri`
- we now stopped creating `extractionError` notifications for `!isErrorWithMessage(e)` case

It would also be a good idea to make this particular logic more local to what can actually throw a syntax error ([this call](https://github.com/statelyai/xstate-tools/blob/417d9d28d1215078726879bd3ac6e3d52c3be236/apps/extension/server/src/server.ts#L196)). Right now we catch just anything from the whole function and we treat everything as "syntax error". To focus on the improvement at hand though that perhaps should be refactored in a separate PR in the future.

apps/extension/server/src/server.ts
```diff
@@ -308,9 +309,12 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
...
+      if (previouslyCachedDocument)
+        previouslyCachedDocument.syntaxError = e.message;
```
Andarist, line 314:
```suggestion
      if (previouslyCachedDocument) {
        previouslyCachedDocument.syntaxError = e.message;
      }
```

apps/extension/server/src/server.ts
mellson: Good observations! I've implemented your suggestions and verified that it works as expected. I agree we should make it more local in a follow-up. I've created a task for myself in Linear.

Andarist (approved):

PR #252, opened 2022-11-19
Add support for --silent

gkiely:
Adds support for running xstate-cli in silent mode. Still prints errors.

### DIFF
--- a/apps/cli/src/bin.ts
+++ b/apps/cli/src/bin.ts
@@ -24,7 +24,7 @@ const handleError = (uri: string, e: any) => {
   }
 };
 
-const writeToFiles = async (uriArray: string[]) => {
+const writeToFiles = async (uriArray: string[], silent = false) => {
   /**
    * TODO - implement pretty readout
    */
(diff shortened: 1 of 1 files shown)


apps/cli/src/bin.ts
```diff
@@ -74,7 +74,9 @@ const writeToFiles = async (uriArray: string[]) => {
           filePath: uri,
           event,
         });
-        console.log(`${uri} - success`);
+        if(!silent){
```
davidkpiano, line 77:
```suggestion
        if (!silent) {
```

apps/cli/src/bin.ts
```diff
@@ -108,7 +111,7 @@ program
           if (path.endsWith(".typegen.ts")) {
             return;
           }
-          tasks.push(writeToFiles([path]));
+          tasks.push(writeToFiles([path], opts.silent));
```
davidkpiano, line 114: Might be more scalable to just pass `opts` in here? That way we can avoid multiple arguments in the future.

PR #257, opened 2022-11-23
Implement undo stack that keeps track of deleted AST nodes to bring them back on undo

Andarist:


packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -1469,6 +1513,95 @@ export class MachineExtractResult {
...
+      parser: {
+        // this is a slight hack to defer the work done by `recast.parse`
+        // we don't need to re-parse the file though, as we already have the AST
+        parse: () => this._fileAst,
+      },
+    });
+
+    const recastDefinitionNode = findRecastDefinitionNode(
+      ast,
+      this.machineCallResult.definition!.node,
```
mellson, line 1543: I'm guessing all the added not-null assertions on the machine results is because they can, in fact, never be null, and we don't want to use the type system to express that?

packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -1469,6 +1513,95 @@ export class MachineExtractResult {
...
+      newText: reprinted.slice(
+        machineNode.arguments[0].start!,
+        machineNode.arguments[0].end!,
+      ),
+      deleted: deleted.length ? deleted : undefined,
+    };
+  }
+
+  restore({ deleted }: { deleted: DeletedEntity[] }) {
+    // TODO deduplicate pre and post loop stuff
```
mellson, line 1527: Is this still relevant?

mellson (approved):

packages/machine-extractor/src/MachineExtractResult.ts
Andarist: it is in the sense that it could be done - but I don't intend to refactor this right now as in the ideal world I would just remove lots of this code altogether.

packages/machine-extractor/src/MachineExtractResult.ts
Andarist: Ye, kinda - the codebase is scattered with this right now anyway. I have plans for bigger refactors that would avoid this kind of a thing

apps/extension/server/src/types.ts
```diff
@@ -10,4 +10,5 @@ export interface ExtractionResult {
 export interface CachedDocument {
   documentText: string;
   extractionResults: ExtractionResult[];
+  undoStack: Array<{ deleted: any[] } | undefined>;
```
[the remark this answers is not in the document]
Andarist: Yeah, it's on purpose here because I would have to cast it down the road anyway. The code here doesn't really care about those values. They are just kept here so we can pass them back to the `restore` call. `recast` types are really annoying to work with and I wasn't able to find a type that would represent "any" AST node. From what I understand, the problem is that `recast` handles more than a single AST formats and they are not always compatible

apps/extension/server/src/server.ts
```diff
@@ -234,6 +234,7 @@ async function handleDocumentChange(textDocument: TextDocument): Promise<void> {
     documentsCache.set(textDocument.uri, {
       documentText: text,
       extractionResults,
+      undoStack: previouslyCachedDocument?.undoStack || [],
```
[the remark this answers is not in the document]
Andarist: Old habits never die.

PR #258, opened 2022-11-25
Introduce `newTransitionPath` to `reanchor_transition`

Andarist:


packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -1258,15 +1259,17 @@ export class MachineExtractResult {
               edit.newSourcePath,
             );
 
-            transitionPath = [
-              ...edit.transitionPath.slice(0, -1),
-              getIndexForTransitionPathAppendant(
```
Andarist, line 1263: the problem was that we were always appending here but to enable undo of the `reanchor_transition` we need to insert transitions at their exact previous positions and those don't necessarily have to be at the end of the group

mellson (approved):

PR #259, opened 2022-11-29
Provide `distinctId` to the Editor to keep analytics there

Andarist:
This basically reverts the changes made in [this PR](https://github.com/statelyai/xstate-tools/pull/217) and provides the `vscode.env.machineId` through the injected `window.__params`


apps/extension/client/src/initiateEditor.ts
```diff
@@ -91,6 +91,7 @@ async function getWebviewHtml(
     config,
     layoutString,
     implementations,
+    machineId: `vscode:${vscode.env.machineId}`,
```
mellson, line 94: nit: It's a bit confusing that vs code calls this `machineId` since we already have this in our terminology. Should we call it `deviceId` on our end?

mellson (approved):

apps/extension/client/src/initiateEditor.ts
Andarist: Ah, I've missed that comment when taking a look at our internal PR in isolation. Yeah... this makes sense. Gonna adjust it in a second.

PR #262, opened 2022-12-01
Clear `undoStack` when the editor gets closed

Andarist:


apps/extension/server/src/server.ts
```diff
@@ -805,11 +805,21 @@ connection.onRequest('getNodePosition', ({ path }) => {
...
 
 connection.onRequest('clearDisplayedMachine', () => {
+  if (!displayedMachine) {
+    throw new Error('There is no displayed machine to clear');
+  }
+  const cachedDocument = documentsCache.get(displayedMachine.uri);
+  if (!cachedDocument) {
+    throw new Error(`Document for the displayed machine couldn't be found`);
+  }
+  cachedDocument.undoStack = [];
```
Andarist, line 822: This avoids a memory leak. And a potential data leak between "editing sessions" - this shouldn't happen in practice as we shouldn't dispatch undo edits from the Editor when there is nothing to undo there, but this acts as an extra safety measure 😉

mellson (approved):

Issue #265, opened 2022-12-06
version 1.11.0 typegen broken on windows, works on mac.


theslantedroom:
Good day, just upgraded to 1.11.0 on my windows machine and had the following issue:

the typegen file apears to generate incorrectly.  The result was typescript errors in the app.

![xstate vscode 1 11 0 windows](https://user-images.githubusercontent.com/73557008/206046851-66c6beb4-eb04-47d0-8388-eb9bb5fbcd44.png)

You will notice the missingImplementations and eventsCausingServices were generated differently.
This is not the case on mac / Xstate vs code extention 1.11.0

Temp fix:  downgrading to 1.10.0 and I am back in business.

Andarist:
Hmm, this is quite weird - especially that you mention this being specific to the operating system. Any chance that you would be able to provide a repro case? I have no idea how to start investigating this without one

theslantedroom:
I may be able to. Let me see what I can do.  Additionally, I confirmed this to be the case on two different PCs.

Andarist:
I think I've found the issue and I'm working on the fix [here](https://github.com/statelyai/xstate-tools/pull/267). However, this problem doesn't quite relate to the OS - so there is still a minimal chance that your issue here is different somehow.

theslantedroom:
https://user-images.githubusercontent.com/73557008/206540686-6e1dbf48-32e6-46d7-9125-68b7fc555152.mp4

I have created a simple repo for you with a working (bug exists) example 

https://github.com/theslantedroom/XstateTypeGenBug1.11.0

nomoon:
I believe I've just had this occur on Mac, using the latest VS Code and XState extension. I opened a previously-working machine file, I guess the Typegen re-processed, and suddenly I have  errors, with this in my typegen:

```typescript
  missingImplementations: {
    actions: "j6k8t"; // <-- that's not a real action!
    delays: never;
    guards: never;
    services: never;
  };
```

Glad to hear a fix is in the works!

PR #267, opened 2022-12-08
Do not include hashed IDs of inline implementations in the typegen data

Andarist:
closes https://github.com/statelyai/xstate-tools/issues/265


packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -199,28 +202,25 @@ export const getActionConfig = (
...
+          type: 'anonymous',
+        });
+        return;
+      case opts?.hashInlineImplementations:
+        actions.push({
+          type: action.inlineDeclarationId,
+        });
+        return;
+      case opts?.stringifyInlineImplementations:
         actions.push(
```
mellson, line 220: Should this be guarded, just in case?

```ts
if (opts && action.node.start && action.node.end) {
  actions.push(
    opts.fileContent.slice(action.node.start, action.node.end),
  );
}
```

mellson (approved):
The code looks good. Did you test and verify against the example repo from #265 ?

packages/machine-extractor/src/toMachineConfig.ts
Andarist: Not really - those properties are truly always there in our case. The types just don't reflect that as they are too close to what Babel exposes. I plan to improve this in the future but since they are always there it would obscure the code a lot if we'd guard against this all over the place.

Andarist:
> Did you test and verify against the example repo from https://github.com/statelyai/xstate-tools/issues/265 ?

Yeah, I tested it against one of the repros there.

PR #271, opened 2022-12-09
Fixed issues with some properties not being removable

Andarist:
Those edits were correctly exchanged between the Studio and the extension... but they were malformed within the extension when passing them to the language server.


mellson (approved):
LGTM 😊

packages/machine-extractor/tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "outDir": "out",
     "rootDir": "src",
     "noEmit": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "exactOptionalPropertyTypes": true
```
davidkpiano, line 13: Is this in general a good thing to enable? Wouldn't mind having this in the studio

davidkpiano (approved):

packages/machine-extractor/tsconfig.json
Andarist: I think so. However, this would require a lot more work than in this repo - since the Studio is a much bigger project and it also integrates with way more packages.

Issue #269, opened 2022-12-08
Bug: VSCode extension "XState Editor" does not write back to the file


viglucci:
### Description

When opening the XState Editor to edit a machine via the "Open Visual Editor" action, the editor does not seem to initialize properly. Changes made in the visual editor are not persisted back to the file containing the machine definition. Additionally, the xstate layout comment is no longer added above the machine definition on new machines or edited when editing existing machines.

### Expected result

Changes made in the visual editor are persisted back to the file containing the machine definition.

### Actual result

Changes made in the visual editor are not persisted back to the file containing the machine definition.

### Reproduction

n/a

### Additional context

**Repro**: I did not provide repro as not sure how to provide a repro of the plugin. Happy to provide any logs or other diagnostic information if it exists.

**Platform**: Windows 10

```
$ node -v
v18.12.1
```

vscode: `1.74.0`

extension version: `1.11.0`

Potentially related to #229 . This seemed to be working fine before `1.11.0`.

viglucci:
Regarding this being introduced with `1.11.0`, I just manually installed `v1.10.0` via `vsix` and the issue appears to be resolved.

```
code --force --install-extension .\statelyai.stately-vscode-1.10.0.vsix
```

viglucci:
Hey @Andarist, thanks for the reply. I can't share the entire fire where I first encountered this issue, however, I can reproduce it with even the most simple of machines.

example:

```ts
import { createMachine } from "xstate";

export default createMachine({
    id: "example",
});
```

Please see attached where I demonstrate attempting to edit and existing complex machine, and then the second example where I attempt to create a brand new simple machine.

https://user-images.githubusercontent.com/6305490/206603496-37681fbc-12e6-4bbb-a122-5a5cd4125d10.mp4

https://user-images.githubusercontent.com/6305490/206603600-bbe1f318-f765-47cf-bc85-3447edd976f7.mp4

davidkpiano:
Thanks for the report, @viglucci. I can also reproduce this. There is a workaround where if you declare a normal `const machine = createMachine({...})`, it works as expected:

https://user-images.githubusercontent.com/1093738/206616119-f16dc7e0-3b77-4ac5-80cb-5877a84eda10.mp4

We will look into this tomorrow.

PR #272, opened 2022-12-09
Improved displayed error when the displayed machine index can no longer be found in the source code

Andarist:
I noticed the problem on David's video here: https://github.com/statelyai/xstate-tools/issues/269#issuecomment-1343789263

The error alone, when commenting out a machine, wasn't that bad - but a bigger problem was that the error wasn't going away when uncommenting the machine.


apps/extension/server/src/utils.ts
```diff
@@ -11,6 +11,18 @@ export const isErrorWithMessage = (
   'message' in error &&
   typeof (error as Record<string, unknown>).message === 'string';
 
+export const getErrorMessage = (error: unknown): string => {
```
mellson, line 14: If we want we can reuse the `isErrorWithMessage` helper here and do:
```ts
const getErrorMessage = (error: unknown): string => isErrorWithMessage(error) ? error.message : 'Unknown error';
```

If we don't want to do that we should just remove `isErrorWithMessage` entirely as it is not used anymore.

mellson (approved):

apps/extension/server/src/utils.ts
Andarist: Ah ye, my intention was to remove it - i forgot to do it. Thanks for noticing 👍

PR #273, opened 2022-12-09
Fixed an occasional crash when applying machine edits to machines directly exported from a file

Andarist:
fixes https://github.com/statelyai/xstate-tools/issues/269

### DIFF
--- a/packages/machine-extractor/src/MachineExtractResult.ts
+++ b/packages/machine-extractor/src/MachineExtractResult.ts
@@ -708,8 +708,9 @@ export class MachineExtractResult {
   };
 
   modify(edits: Array<MachineEdit>) {
-    // we need to read it before calling `recast.parse` as that mutates the AST and removes comments
+    // we need to read it before calling `recast.parse` as that mutates the AST and removes comments and sometimes also locations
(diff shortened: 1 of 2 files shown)


packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -708,8 +708,9 @@ export class MachineExtractResult {
   };
 
   modify(edits: Array<MachineEdit>) {
-    // we need to read it before calling `recast.parse` as that mutates the AST and removes comments
+    // we need to read it before calling `recast.parse` as that mutates the AST and removes comments and sometimes also locations
     const existingLayoutComment = this.getLayoutComment();
+    const existingMachineNodeLoc = this.machineCallResult.node.loc!;
```
Andarist, line 713: I don't know why this case was special - but I've confirmed that `recast` is sometimes removing `.loc` from nodes. So we have to save the existing `loc` here - before we pass the whole thing to `recast` 🙈

mellson: Nice find 🕵🏻

mellson (approved):

PR #305, opened 2023-02-15
Prefer inserting an inner layout comment

Andarist:
STA-2082

The layout string will only be inserted inside the machine when there is no existing layout machine - and of course, once it's inside it will be updated there.

For the time being, I decided to avoid "migrating" the old placement to the new one. This would create big diffs in projects as the machine config is almost guaranteed to be dedented after a such change and that could get in people's way during code reviews etc. We could consider making this configurable or something but let's start simple.

### DIFF
--- a/packages/machine-extractor/src/MachineExtractResult.ts
+++ b/packages/machine-extractor/src/MachineExtractResult.ts
@@ -1537,17 +1537,20 @@ export class MachineExtractResult {
               };
               break;
             }
+            const indentation = consumeIndentationToNodeAtIndex(
+              this._fileContent,
+              definitionNode.properties[0]?.start,
+            );
             const insidePosition = {
               line: definitionNode.loc!.start.line - 1,
               column: definitionNode.loc!.start.column + 1,
               index: definitionNode.start! + 1,
             } as const;
-
             layoutEdit = {
               // this is used as a replace but it could be a simpler~ insertion
               type: 'replace',
               range: [insidePosition, insidePosition],
-              newText: `\n/** @xstate-layout ${edit.layoutString} */`,
+              newText: `\n${indentation}/** @xstate-layout ${edit.layoutString} */`,
             };
             break;
           }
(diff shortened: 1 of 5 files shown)


packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -1490,16 +1521,32 @@ export class MachineExtractResult {
...
+              column: definitionNode.loc!.start.column + 1,
+              index: definitionNode.start! + 1,
             } as const;
 
             layoutEdit = {
               // this is used as a replace but it could be a simpler~ insertion
               type: 'replace',
-              range: [calleePosition, calleePosition],
+              range: [insidePosition, insidePosition],
               newText: `\n/** @xstate-layout ${edit.layoutString} */\n`,
```
mellson, line 1550:
```suggestion
              newText: `\n/** @xstate-layout ${edit.layoutString} */`,
```
nit: I prefer not to have an empty line between the comment and the next thing. But I understand it's subjective, so pick whichever you prefer. I'm only suggesting this here, but if it's changed, I guess it should also be changed in the block above this starting line 1536.

**Indentation**
Right now, we don't ignore the existing indentation when we insert. Is it too naive to use something like this:
```ts
...
const indent = (definitionNode.loc as any).lines.infos[
  definitionNode.loc!.start.line
].indent;
const indentString = ' '.repeat(indent); // not sure this is fully compatible with tabs for indent, but I guess that would be handled by the user's formatter if it is a problem

layoutEdit = {
  // this is used as a replace but it could be a simpler~ insertion
  type: 'replace',
  range: [insidePosition, insidePosition],
  newText: `\n${indentString}/** @xstate-layout ${edit.layoutString} */`,
};
...
```

mellson (approved):

packages/machine-extractor/src/MachineExtractResult.ts
davidkpiano: Would this insert an empty line, or would it merely make sure that the machine config code is on the next line?

mellson: In the test I did it created a new empty line after the layout comment

davidkpiano:
Fixes STA-2082 (come on, Linear, do your thing)

davidkpiano:
Is this good to go in?

Andarist:
@davidkpiano not yet, I only need to fix a small issue with a redundant new line - gonna handle that tomorrow and land this

mellson (approved):
This is working brilliantly, with indentation and all - great job 👏🏻👏🏻👏🏻

PR #332, opened 2023-04-27
initialize @xstate/codemods package

### LINKED ISSUE #321: v5 migration codemod
This issue is for building a codemod that migrates a codebase from v4 to v5. The codemod would analyze the abstract syntax tree (AST) of the codebase in order to identify syntax changes that need to be made and apply those changes.

You can find the current (as of 4/14) migration guide here: https://docs-git-davidkpiano-v5-outline-statelyai.vercel.app/xstate-v5/migration

Here's a table of changes to target:

|v4|v5|Changes to make|Comments/Questions|
|-|-|-|-|
|`Machine`|`createMachine`|<ul><li>replace `Machine` identifier with `createMachine`</li><li>replace `Machine` import with `createMachine`</li></ul>||
|`guard`|`cond`|<ul><li>replace `guard` keys with `cond`</li></ul>||
|`(context, event)`|`({context, event})`|<ul><li>wrap arguments in object syntax</li></ul>||
|`actor.onTransition`|`actor.subscribe`| ? | do they have the same interface? |
|`send`|`raise`/`sendTo`|?|is there a way to tell from `send` usage which new fn should be used instead?|
|`internal: false`| `reenter: true` | <ul><li>if `internal: true`, remove property</li><li>if `internal: false`, remove and replace with new `reenter: true` property</li></ul> ||
|`interpret(machine).start(state)`|`interpret(machine, {state}).start()`|<ul><li>remove `state` argument from `start` call</li><li>create/update 2nd `interpret` argument object with `state` property</li></ul>||
|`actor.send('event')`|`actor.send({type: 'event'})`|<ul><li>remove string argument</li><li>create object argument with `type: 'event'` property</li></ul>||
|null event (`''`)|`always`|<ul><li>replace `''` key with `always`</li></ul>||
| `createMachine(config, options)` | `createMachine(config).provide(options)` | <ul><li>remove 2nd argument from `createMachine`</li><li>append call to `.provide` with `options` as its argument</li></ul> | |
| `machine.withConfig(options, context)` | `machine.provide({…options, context})` | <ul><li>remove `.withConfig` call</li><li>add `.provide` call</li><li>if `context` is passed <ul><li>add empty object argument to `provide` call</li><li>spread `options` in object argument</li><li>add `context` property to object argument</li></ul></li><li>if `context` is not passed, add `options` as argument to `provide`</li></ul> | need to handle cases where both `.withConfig` and `.withContext` were called |
| `machine.withContext(context)` | `machine.provide({context})` | <ul><li>remove `.withContext` call</li><li>add `.provide` call</li><li>add object with `context` property as argument to `provide`</li></ul> | need to handle cases where both `.withConfig` and `.withContext` were called |
|`predictableActionArguments: true`| - |<ul><li>remove `predictableActionArguments` property</li></ul>||
| `strict: true` | - | <ul><li>remove `strict` property</li></ul> | |
| `state.history` | - | <ul><li>remove `state.history` property</li></ul> | is there anything else we could do here? maybe notify the user that `state.history` no longer exists and where it's being used so they can adjust their code? |
| `actor.batch([...])` | removed | <ul><li>remove `actor.batch` call</li><li>create an individual `actor.send` call for each event in the array</li></ul> | is this an appropriate solution? |
|`config.schema`|`config.types`|<ul><li>rename `schema` key to `types`</li>| need to make sure this works together with `config.tsTypes` change |
|`config.tsTypes`|`config.types.typegen`|<ul><li>create `types` property with empty object value</li><li>move `tsTypes` property to `types` object</li><li>rename `tsTypes` key to `typegen`</li></ul>| need to make sure this works together with `config.schema` change |

with-heart:
This PR resolves #321 by creating the `@xstate/codemods` package and implementing TypeScript transformers for each of the v4-to-v5 changes.

This is currently just a draft as we figure out the ideal way to implement the codemod and handle all of the necessary changes.


with-heart:
The initial impl for `machine-to-create-machine` is super naive. We're traversing for each node in the tree that needs to be renamed and renaming it, which works but it's a *lot* of effort compared to what we could do with the ts language server or [`ts-morph`](https://github.com/dsherret/ts-morph). Going to investigate those alternatives next I think

packages/codemods/src/v4-to-v5/transformers/machine-to-create-machine.ts
```diff
@@ -0,0 +1,48 @@
...
+
+      if (isMachinePropertyAccessExpression(node)) {
+        return factory.updatePropertyAccessExpression(
+          node,
+          node.expression,
+          factory.createIdentifier('createMachine'),
+        );
+      }
+
+      return ts.visitEachChild(node, visit, context);
```
with-heart, line 46: `visitEachChild` calls `visit` on each of the node's child nodes. This allows us to make `visit` _recursive_.

For each node, we either return an updated version of it (if one of the `if` blocks apply for that node) or we visit each of the node's children and return them. This allows us to traverse through (nearly) every node in the tree, just by returning `ts.visitEachChild`!

packages/codemods/src/v4-to-v5/predicates.ts
with-heart, line 1: _Predicates_ are functions that take a node and return a boolean. They're useful when traversing the tree with visitor functions because they allow us to select specific types of nodes that we want to operate on.

TypeScript itself exports many predicate fns—one for each type of node.

In this module, we combine the ts predicates with a few additional checks of node properties to create functions that allow us to select each type/shape of node our codemod needs in order to do its thing.

with-heart: The other cool thing about predicates is that _all_ of the TypeScript predicates and _some_ of these predicates return a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) (the `node is ts.XYZ` syntax). 

A type predicate _narrows_ the type of its argument, meaning that when used with a conditional check, a less specific type (`ts.Node`) is _narrowed_ to a more specific type (`ts.XYZ`).

Here's an example:

```ts
import ts from 'typescript'

declare const node: ts.Node

ts.text // type error (because not all nodes have the text property)

if (ts.isIdentifier(node)) {
  // node is now of type ts.Identifier
  ts.text // totally fine!
}
```

with-heart (commented):
I did my best to give a high-level explanation of how all of this works currently. I could go on and on though so I'm gonna stop here.

If you have any questions about any of it, don't hesitate to add your question as a comment!

packages/codemods/src/v4-to-v5/predicates.ts
with-heart: PS: These predicates will make a lot more sense if you explore the AST from this sample: https://ts-ast-viewer.com/#code/JYWwDg9gTgLgBAbzgWQIYGMAWwB2BTOAXzgDMoIQ4ByADwGcZUY8qAoUSWRFDbfOVHRRFS5SrQZMW7cNHj1GzURWoKpbVugg4GcEL1wEAvDyyGAFAkIBKTdt36z-E8ks27O+I77G4a5gB0aE54btZAA

tivac:
Any further progress planned for this? We're hoping to try out v5 soon but have enough machines that manually porting them all is gonna be pretty time-consuming.

Andarist:
I'd recommend using XState v4 and v5 side by side and migrating to v5 gradually. You want to make sure that the behavior of your system stays unaffected and that the codemod can only migrate some syntax-oriented things, like renaming `cond` to `guard`.

To use both you can add such dependencies:
```ts
"@xstate/react5": "npm:@xstate/react@4.0.1",
"xstate5": "npm:xstate@5.2.0",
```

And if you need an integration package like `@xsttae/react` (or some other) then you need to link it with the correct XState version after you install the packages. You can do this with a script like:
```ts
const fs = require('fs-extra');
const path = require('path');

const rootNodeModules = path.join(__dirname, '..', 'node_modules');

fs.ensureSymlinkSync(
  path.join(rootNodeModules, 'xstate5'),
  path.join(rootNodeModules, '@xstate', 'react5', 'node_modules', 'xstate'),
);
```

davidkpiano:
Added this to the docs: https://stately.ai/docs/migration#how-can-i-use-both-xstate-v4-and-v5

leggomuhgreggo:
Oh wow! This is neat. Any plans to revive this?

davidkpiano:
> Oh wow! This is neat. Any plans to revive this?

I hope to! I've been thinking about this recently - either lint rules and/or a codemod for v5 and future XState versions.

tivac:
Reviving this would be huge for me, the guy who _still_ has a ton of v4 statecharts with no straightforward way to move to v5 without a huge amount of manual work.

with-heart:
I'm working on an rfc for codemods so we can have a solid plan ❤️

PR #334, opened 2023-05-04
Extract inline action expressions according to the new implementations in editor with blocks

farskid:
> No guards support because blocks don't yet treat guards as blocks.

To do:

**Inline Actions:**
- [x] Extract inline `assign` expressions 
- [x] Extract inline `raise` expressions
- [x] Extract inline `log` expressions
- [x] Extract inline `sendTo` expressions
- [x] Extract inline `stop` expressions

**TBD**
- [x] [we don't want to support it right now] ~~Identifier referencing an action or an invocation~~
```ts
const someVariable = assign({})
createMachine({
  entry: [someVariable]
})
```
- [ ] Identifier recursively referencing an action or invocation
```ts
const b = assign({});
const a = b;
createMachine({
  entry: [a]
})
```
- [ ] Identifier wrapping around an action or invocation
```ts
const myAssigner = (...args: Parameters<typeof assign>) => assign(...args);
```


packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,89 @@
...
+  return assignment;
+}
+
+function getLiteralType(value: t.ObjectProperty['value']) {
+  if (
+    t.isNullLiteral(value) ||
+    t.isRegExpLiteral(value) ||
+    t.isTemplateLiteral(value)
+  ) {
+    throw Error('null,regexp or template literans can not have literal type');
```
farskid, line 66: Find out a way to make this general purpose.

packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -234,6 +226,13 @@ export const getActionConfig = (
           }),
         });
         return;
+      case isAssignAction(action):
+        actions.push({
+          type: action.name || `Assignment ${Math.random().toFixed(3)}`,
+          name: 'xstate.assign',
```
farskid, line 232: This should be derived from `actionNode.node` call expression.

with-heart:
Will review this today!

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    // assign(() => {})
+    // assign(function() {})
+    if (
+      t.isArrowFunctionExpression(assigner) ||
+      t.isFunctionExpression(assigner)
+    ) {
+      return {
+        inlineImplementation: {
```
Andarist, line 103: I'd replace this with smth like `{ type: 'expression', value: string } | { type: 'object': properties: Record<string, ValidStuff> }`. This way `inlineImplementation` wouldn't have to be a special key here.

packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -234,6 +226,13 @@ export const getActionConfig = (
           }),
         });
         return;
+      case isAssignAction(action):
+        actions.push({
+          type: action.name || `Assignment ${Math.random().toFixed(3)}`,
```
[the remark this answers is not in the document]
Andarist: We shouldn't auto-generate things here on behalf of the user. This kind of a thing is "lossy" because by the time we end up consuming this, we won't be able to tell apart the provided and the auto-generated strings.

packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -130,109 +132,518 @@ describe('MachineParseResult', () => {
...
-      ).toMatchInlineSnapshot(`
-        [
-          "assign({count: 1})",
-          "assign((ctx, e) => {
+                  count: val + ctx.count,
+                };
+              }",
+            },
+          },
+          "name": "xstate.assign",
```
Andarist, line 188: can a custom action appear here (not a builtin one)? how that would look like?

packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -130,109 +132,518 @@ describe('MachineParseResult', () => {
...
+        },
+        "name": "xstate.assign",
+      }
+    `);
+
+    ['a', 'b', 'c', 'd', 'e'].forEach((prop) => {
+      expect(config?.states?.a?.on?.FOO?.actions.assignment).toHaveProperty(
+        prop,
+      );
     });
```
Andarist, line 230: this bit feels redundant if you are testing all of those in just the next `expect` call

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+  if (t.isStringLiteral(value)) {
+    return 'string';
+  }
+  if (t.isBooleanLiteral(value)) {
+    return 'boolean';
+  }
+  throw Error('Unsupported literal property value');
+}
+
+export const isBuiltinActionWithName = (actionNode: ActionNode, name: string) =>
```
Andarist, line 360:
```suggestion
// TODO: this doesn't yet handle namespace imports and aliased specifiers
export const isBuiltinActionWithName = (actionNode: ActionNode, name: string) =>
```

packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -130,109 +132,518 @@ describe('MachineParseResult', () => {
...
+              "id": undefined,
+            },
+          },
+          "name": "xstate.sendTo",
+        },
+      ]
+    `);
+  });
+
+  it('Should extract stop expressions', () => {
```
Andarist, line 600: I'd prefer to have specific tests for those things, further down the road it's much better to see a failing test case title "extract stop action (string argument)" than to see "extract stop actions (all variants)". It also helps debugging failing test cases because it's easier to focus on a specific test case that might be failing, code paths within the implementation are hit less frequently etc

packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -130,109 +132,518 @@ describe('MachineParseResult', () => {
...
+  });
+
+  it('should extract log action expression', () => {
+    const result = extractMachinesFromFile(`
+    createMachine({
+      initial: "a",
+      context: { count: 0 },
+      entry: [
+        log((ctx, e) => {}),
+        raise('Some string')
```
Andarist, line 326: this test likely wasn't supposed to test `raise`

packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -130,109 +132,518 @@ describe('MachineParseResult', () => {
...
+  });
+
+  it('should extract raise actions', () => {
+    const result = extractMachinesFromFile(`
+    createMachine({
+      initial: "a",
+      context: { count: 0 },
+      entry: [
+        raise({type: 'Some event', foo: 'foo', bar: true, baz: [1,2,3], obj: {prop: {prop2: 2}}}),
+        raise('Some other event')
```
Andarist, line 268: we should also test `raise(() => ({ type: 'FOO' }))`

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+    string,
+    { type: JsonEntryType; value: JsonValue }
+  > = {};
+  eventObject.properties.forEach(prop => {
+    if (t.isObjectProperty(prop)) {
+      if (t.isIdentifier(prop.key)) {
+        if (t.isLiteral(prop.value)) {
+          if (
+            t.isRegExpLiteral(prop.value) ||
+            t.isTemplateLiteral(prop.value) ||
```
Andarist, line 379: expression-less template literals, like this one:
```ts
`just a string`
```
should be treated as a string, not as an expression

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+}
+
+function getLiteralType(value: t.ObjectProperty['value']) {
+  if (
+    t.isNullLiteral(value) ||
+    t.isRegExpLiteral(value) ||
+    t.isTemplateLiteral(value)
+  ) {
+    throw Error('null,regexp or template literals can not have literal type');
+  }
```
Andarist, line 347: what does it mean? I see those being often handled before calling `getLiteralType` and I'm not sure what's the distinction

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+  if (t.isNumericLiteral(value)) {
```
Andarist, line 348: bigint could be handled here as well

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+        {
+          type: JsonEntryType;
+          value: JsonValue;
+        }
+      >
+    | { type: 'expression'; value: string };
+  actorRef: {
+    type: JsonEntryType;
+    value: JsonValue;
+  };
```
Andarist, line 209: just reordered those things to match the order of arguments:
```suggestion
  actorRef: {
    type: JsonEntryType;
    value: JsonValue;
  };
  event:
    | Record<
        string,
        {
          type: JsonEntryType;
          value: JsonValue;
        }
      >
    | { type: 'expression'; value: string };
```

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+    // Actor
+    // sendTo('actorName')
+    if (t.isStringLiteral(arg1)) {
+      actorRef = { type: 'string', value: arg1.value };
+    }
+    // sendTo((ctx, e) => actorRef)
+    else if (
+      t.isArrowFunctionExpression(arg1) ||
+      t.isFunctionExpression(arg1)
+    ) {
```
Andarist, line 248: why do we special case functions here? what about `sendTo(someActorRef, { type: 'FOO' })`?

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+        id = {
+          type: getLiteralType(foundId.value),
+          value: foundId.value.value,
+        };
+      }
+    }
+
+    return {
+      event: eventObject,
+      actorRef: actorRef!,
```
Andarist, line 329:
```suggestion
      actorRef: actorRef!,
      event: eventObject,
```

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+  eventObject: t.ObjectExpression,
+  fileContent: string,
+): Record<string, { type: JsonEntryType; value: JsonValue }> {
+  const extracted: Record<
+    string,
+    { type: JsonEntryType; value: JsonValue }
+  > = {};
+  eventObject.properties.forEach(prop => {
+    if (t.isObjectProperty(prop)) {
+      if (t.isIdentifier(prop.key)) {
```
Andarist, line 375: You should throw/warn/do something about unrecognized patterns - fail fast instead of propagating incomplete/broken results forward.

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+    }
+  > = {};
+  if (t.isCallExpression(node)) {
+    const assigner = node.arguments[0];
+
+    // assign({})
+    if (t.isObjectExpression(assigner)) {
+      assigner.properties.forEach(prop => {
+        if (t.isObjectProperty(prop)) {
+          if (t.isIdentifier(prop.key)) {
```
Andarist, line 44: what about spreads and other unrecognized patterns?

packages/machine-extractor/src/extractAction.ts
```diff
@@ -0,0 +1,411 @@
...
+          value: fileContent.slice(assigner.start!, assigner.end!),
+        },
+      };
+    }
+  }
+
+  throw Error(`Unsupported assignment`);
+}
+
+export function extractRaisedEvent(
```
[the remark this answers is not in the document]
farskid: We discussed offline and agreed to support it later.

packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -234,6 +234,37 @@ export const getActionConfig = (
...
+      // Todo: think about error reporting and how to handle invalid actions such as raise(2)
+      case isBuiltinActionWithName(action, 'assign'):
+        actions.push({
+          type: 'xstate.assign',
+          assignment: extractAssignAction(action, opts!.fileContent),
+        });
+        return;
+      case isBuiltinActionWithName(action, 'raise'):
+        actions.push({
+          type: 'xstate.raise',
```
farskid, line 246: Add `name` again.

Issue #340, opened 2023-05-09
[Bug] @XState/cli@0.5.0 is generating empty typegen files


SimeonC:
When running `xstate typegen "./(cypress|src)/**/*.ts?(x)"` XState machines that don't have `tsTypes` still generate empty `*.typegen.ts` files with just a comment.

https://github.com/statelyai/xstate-tools/blob/e195dfe65479cdec5ccb1e7d42656b4e8d6b199c/packages/shared/src/writeToTypegenFile.ts#L15-L18

This check didn't get moved in [this commit](https://github.com/statelyai/xstate-tools/commit/87bf22aeb47e1e08edad373c218ca2a0abd0f9d7#). This causes errors in typescript if running with `--isolatedModules`, not to mention a lot of random unnecessary files.

SimeonC:
Thanks for the quick fix. New version is working correctly!

PR #341, opened 2023-05-09
Fixed an issue with typegen files being sometimes generated by the CLI when they shouldn't be

Andarist:
Essentially the same change as [here](https://github.com/statelyai/xstate-tools/pull/327/files). At that time, `@xstate/cli` didn't require any changes but later I've done some refactors in https://github.com/statelyai/xstate-tools/pull/328 and missed doing this here.

fixes https://github.com/statelyai/xstate-tools/issues/340

### DIFF
--- a/.changeset/kind-meals-jog.md
+++ b/.changeset/kind-meals-jog.md
@@ -1,0 +1,5 @@
+---
+'@xstate/cli': patch
+---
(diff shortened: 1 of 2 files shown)


mellson (approved):
Looks good!

PR #344, opened 2023-05-25
Extract named implementations

farskid:
Stacks on top of https://github.com/statelyai/xstate-tools/pull/334

Todo:
- [x] Extract action implementation to implementation interface according to what Studio blocks expect
- [ ] Extracted implementations should be a part of end machine config
- [ ] Trace imported references in the implementation and add them to `imports` and `dependentPackages` and copy an import declaration for them inside the `fullCode`.


farskid:
This PR is obsolete by the latest changes to Stately Studio and Machine extractor.

PR #349, opened 2023-06-10
enable linting

with-heart:
This PR upgrades and configures `eslint` for the repo, resolves lint errors in packages files, and adds a `lint` step to the `ci` actions workflow.

I made this PR for two reasons:

1. To help encourage healthier practices moving forward
2. To draw attention to how unsafe some of the most critical code in the project is (like `packages/machine-extractor/src/MachineExtractResult.ts`)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,20 +1,60 @@
-/**@type {import('eslint').Linter.Config} */
-// eslint-disable-next-line no-undef
-module.exports = {
-  // root: true,
-  // parser: '@typescript-eslint/parser',
-  // plugins: [
-  // 	'@typescript-eslint',
-  // ],
-  // extends: [
-  // 	'eslint:recommended',
-  // 	'plugin:@typescript-eslint/recommended',
-  // ],
-  // rules: {
-  // 	'semi': [2, "always"],
-  // 	'@typescript-eslint/no-unused-vars': 0,
-  // 	'@typescript-eslint/no-explicit-any': 0,
-  // 	'@typescript-eslint/explicit-module-boundary-types': 0,
-  // 	'@typescript-eslint/no-non-null-assertion': 0,
-  // }
+/** @type {import('eslint').Linter.Config} */
+const config = {
+  extends: [
+    'eslint:recommended',
+    'plugin:eslint-comments/recommended',
+    'plugin:@typescript-eslint/recommended',
+  ],
+  env: {
+    browser: true,
+    node: true,
+  },
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: [
+      './tsconfig.eslint.json',
+      './apps/**/tsconfig.json',
+      './packages/*/tsconfig.json',
+    ],
+  },
+  plugins: ['@typescript-eslint'],
+  root: true,
+  rules: {
+    'eslint-comments/no-unused-disable': 'error',
+  },
+  overrides: [
+    // override for all typescript files
+    {
+      files: ['**/*.{ts,tsx}'],
+      extends: [
+        // enable type-aware linting rules
+        'plugin:@typescript-eslint/recommended-requiring-type-checking',
+      ],
+    },
+    // override for workspace config/script files
+    {
+      files: ['*.js', 'scripts/*.js'],
+      rules: {
+        // disable `require` errors
+        '@typescript-eslint/no-var-requires': 'off',
+      },
+    },
+    // override for test files
+    {
+      files: ['**/__tests__/**/*.{ts,tsx}', '**/*.{spec,test}.{ts,tsx}'],
+      extends: ['plugin:jest/recommended'],
+      plugins: ['jest'],
+      rules: {
+        '@typescript-eslint/no-empty-function': 'off',
+        '@typescript-eslint/no-non-null-assertion': 'off',
+        '@typescript-eslint/no-unsafe-call': 'warn',
+        '@typescript-eslint/no-unsafe-argument': 'warn',
+        '@typescript-eslint/no-unsafe-assignment': 'warn',
+        '@typescript-eslint/no-unsafe-member-access': 'warn',
+      },
+    },
+  ],
 };
+
+module.exports = config;
--- a/apps/cli/src/__tests__/__examples__/bug-report-1.ts
+++ b/apps/cli/src/__tests__/__examples__/bug-report-1.ts
@@ -1,6 +1,6 @@
 import { assign, createMachine } from 'xstate';
 
-type Data = {};
+type Data = unknown;
 
 type Context = {
   data: Data[];
@@ -73,9 +73,11 @@ const machine = createMachine(
        * This service will now type error if it
        * returns anything other than { id: string }
        */
+      // eslint-disable-next-line @typescript-eslint/require-await
       pollData: async () => {
         return [];
       },
+      // eslint-disable-next-line @typescript-eslint/require-await
       getData: async () => {
         return [];
       },
--- a/apps/cli/src/bin.ts
+++ b/apps/cli/src/bin.ts
@@ -18,7 +18,7 @@ async function removeFile(filePath: string) {
   try {
     await fs.unlink(filePath);
   } catch (e: any) {
-    if (e?.code === 'ENOENT') {
+    if ((e as { code: string })?.code === 'ENOENT') {
       return;
     }
     throw e;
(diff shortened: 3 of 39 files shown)


with-heart:
Switched to draft bc I'm gonna try to get linting configured for tests and want to review the `eslint-disable-next-lint` comments I added to see if there aren't problems I can just fix directly instead.

.eslintrc.js
```diff
@@ -1,20 +1,51 @@
...
+    browser: true,
+    node: true,
+  },
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: [
+      './tsconfig.eslint.json',
+      './apps/**/tsconfig.json',
+      './packages/*/tsconfig.json',
```
with-heart, line 18: Configure TypeScript project support by pointing to the `tsconfig.json` files in the workspace. 

`tsconfig.eslint.json` will serve as the ts project for non-package files in the workspace, like configs/scripts and test files.

.eslintrc.js
```diff
@@ -1,20 +1,60 @@
...
-  // ],
-  // rules: {
-  // 	'semi': [2, "always"],
-  // 	'@typescript-eslint/no-unused-vars': 0,
-  // 	'@typescript-eslint/no-explicit-any': 0,
-  // 	'@typescript-eslint/explicit-module-boundary-types': 0,
-  // 	'@typescript-eslint/no-non-null-assertion': 0,
-  // }
+/** @type {import('eslint').Linter.Config} */
+const config = {
```
with-heart, line 2: We define `config` as a variable so we can attach the `Linter.Config` type using jsdoc. This is supposed to work with `module.exports` but doesn't due to https://github.com/microsoft/TypeScript/issues/47107.

.eslintrc.js
```diff
@@ -1,20 +1,60 @@
...
+  extends: [
+    'eslint:recommended',
+    'plugin:eslint-comments/recommended',
+    'plugin:@typescript-eslint/recommended',
+  ],
```
with-heart, line 7: `extends` sets the baseline rules that are enabled for all files.

We're using the rules recommended by `eslint`, `@typescript-eslint` (which disables and replaces some of the rules recommended by `eslint`), and [`eslint-plugin-eslint-comments`](https://mysticatea.github.io/eslint-plugin-eslint-comments/).

.eslintrc.js
```diff
@@ -1,20 +1,60 @@
...
+  },
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: [
+      './tsconfig.eslint.json',
+      './apps/**/tsconfig.json',
+      './packages/*/tsconfig.json',
+    ],
+  },
```
with-heart, line 20: `@typescript-eslint` allows us to configure linting to be handled on a per-project basis. In this case, we set `projects` as a root `tsconfig.eslint.json` file (which will cover all non-package files like configs, scripts, and test files) and and the `tsconfig.json` files from each of our packages.

.eslintrc.js
```diff
@@ -1,20 +1,60 @@
...
+      plugins: ['jest'],
+      rules: {
+        '@typescript-eslint/no-empty-function': 'off',
+        '@typescript-eslint/no-non-null-assertion': 'off',
+        '@typescript-eslint/no-unsafe-call': 'warn',
+        '@typescript-eslint/no-unsafe-argument': 'warn',
+        '@typescript-eslint/no-unsafe-assignment': 'warn',
+        '@typescript-eslint/no-unsafe-member-access': 'warn',
+      },
+    },
```
with-heart, line 56: This override applies to our test files.

First, we enable [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest) plugin and extend from its recommended rules.

Then we disable or change some `@typescript-eslint` errors to warnings because it doesn't necessarily make sense for tests for be entirely type-safe.

tsconfig.eslint.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
+    "types": ["@types/node", "jest"],
+    "noEmit": true,
+    "allowJs": true
+  },
+  "extends": "./tsconfig.json",
+  "include": [".eslintrc.js", "**/*.js", "**/__tests__/**/*.ts"]
```
with-heart, line 8: This new `tsconfig` file is used to add linting coverage to all of our non-package files: `js` configs + scripts and our test files. It just extends from the root `tsconfig.json` and overrides a few compiler options.

apps/cli/src/__tests__/__examples__/bug-report-1.ts
```diff
@@ -1,6 +1,6 @@
 import { assign, createMachine } from 'xstate';
 
-type Data = {};
+type Data = unknown;
```
with-heart, line 3: With the configuration out of the way, the bulk of the work here was resolving the lint errors that followed.

Some errors I was able to resolve with small changes like this.

apps/cli/src/__tests__/__examples__/bug-report-1.ts
```diff
@@ -73,9 +73,11 @@ const machine = createMachine(
        * This service will now type error if it
        * returns anything other than { id: string }
        */
+      // eslint-disable-next-line @typescript-eslint/require-await
```
with-heart, line 76: Some errors are resolved by disabling them because we know what we're doing, like intentionally disabling `@typescript-eslint/require-await` for this `async` function which reproduces a bug report.

packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -1334,11 +1335,14 @@ export class MachineExtractResult {
 
           if (n.ArrayExpression.check(transitionProp.value)) {
             if (transitionProp.value.elements.length === 1) {
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
with-heart, line 1338: But a lot of errors are disabled because that's currently the only way to make them stop without a lot of effort, like this unsafe assignment where we cast `minifiedTransition as any`.

To me, these cases are the main point of this PR. We're frequently using unsafe code in important parts of the codebase and that makes it difficult to maintain.

Manually resolving these errors is probably worth spending a good amount of time on, and now we can easily track them down by removing the comments associated with them.

apps/cli/src/bin.ts
```diff
@@ -51,7 +53,7 @@ const writeToTypegenFile = async (
   await fs.writeFile(
     typegenUri,
     // // Prettier v3 returns a promise
-    await prettierInstance.format(getTypegenOutput(types), {
```
Andarist, line 54: this `await` here was intentional (as mentioned in the comment :P)

packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -2781,7 +2812,7 @@ function consumeIndentationToNodeAtIndex(
     return '';
   }
   let indentation = '';
-  let i = index;
+  // eslint-disable-next-line no-constant-condition
```
Andarist, line 2815: Let's allow loops with `checkLoops: false`: https://eslint.org/docs/latest/rules/no-constant-condition#checkloops

packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -2531,17 +2558,21 @@ function removeTransitionAtPath(
...
           removeProperty(obj, propPath[0]);
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
           return removed as any;
         }
         case 1: {
           const prop = obj.properties[index];
           n.ObjectProperty.assert(prop);
           // array elements should be expressions and property values should be expressions too
           // so this should always be valid in practice
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
Andarist, line 2569: is there a way to relax this rule to ignore lines with explicit `as any`?

packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -747,13 +747,14 @@ export class MachineExtractResult {
     // so there is a risk that modifying multiple machines in a single file would lead to problems
     // however, we never modify multiple machines based on the same file content so this is somewhat safe
     // each modification updates the AST, that is printed and the file is re-parsed, so the next modification sees the next AST
-    const ast: RecastFile = recast.parse(this._fileContent, {
+    this._fileContent;
```
Andarist, line 750: Some leftover here, perhaps?
```suggestion
```

apps/extension/client/src/initiateEditor.ts
```diff
@@ -433,7 +449,9 @@ const machine = createMachine(
             },
           );
 
+          // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
Andarist, line 452: Ugh, I wish this wouldn't be flagged as a problem :v

apps/cli/src/bin.ts
```diff
@@ -63,7 +65,9 @@ const allSettled: typeof Promise.allSettled = (promises: Promise<any>[]) =>
   Promise.all(
     promises.map((promise) =>
       promise.then(
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
Andarist, line 68: hm, what's unsafe here? 🤔

PR #353, opened 2023-06-17
fix: prettier run on newText

### LINKED ISSUE #351: Prettier not run on newText
## Bug

Prettier is only run on typegen file, and not the edits themselves. This causes double quotes when a custom prettier config could change them to single quotes.

dylanarmstrong:
Determine quote style from first found import statement in machine file, then use that for quoting the typegen import.

Closes #351

### DIFF
--- a/apps/cli/src/bin.ts
+++ b/apps/cli/src/bin.ts
@@ -42,20 +42,26 @@ function getPrettierInstance(cwd: string): typeof import('prettier') {
   }
 }
 
+// Used to prettify text before writing
+const prettify = async (
+  uri: string,
+  text: string,
+  { cwd }: { cwd: string },
+) => {
+  const prettierInstance = getPrettierInstance(cwd);
+  return prettierInstance.format(text, {
+    ...(await prettierInstance.resolveConfig(uri)),
+    parser: 'typescript',
+  });
+};
+
 const writeToTypegenFile = async (
   typegenUri: string,
   types: TypegenData[],
   { cwd }: { cwd: string },
 ) => {
-  const prettierInstance = getPrettierInstance(cwd);
-  await fs.writeFile(
-    typegenUri,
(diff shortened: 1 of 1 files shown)


apps/cli/src/bin.ts
```diff
@@ -110,7 +116,11 @@ const writeToFiles = async (uriArray: string[], { cwd }: { cwd: string }) => {
 
         const edits = getTsTypesEdits(types);
         if (edits.length > 0) {
-          const newFile = processFileEdits(fileContents, edits);
+          const newFile = await prettify(
```
Andarist, line 119: Ideally, I'd like to remove the dependency on Prettier altogether. I'd much more prefer to detect the preferred quotes style and just use that within the edit.

Andarist: It's rather a safe bet to assume that there is *some* import statement in the file and we could piggy-back our quotes detection based on that with smth like `/import\s(.|\n)*(['"])/`. 

> This all has a downside of ignoring non-quote prettier features that we may violate like line length and trailing commas.

I understand the concern but we are not quite in the code formatting business. There are multiple formatters available out there and it would be great to use some pretty generic code that could play well with all of them. I think it's fine that sometimes an additional formatting edit might be required after we insert something into the file, the goal is to minimize the chances of that happening.

Prettier is, of course, vastly popular so we could try to special-case it but I think that this should be done by treating is an optional peer dependency~ whereas right now we always fall back to the local version installed with the CLI if we can't load Prettier relative to `cwd`.

dylanarmstrong: I've updated the PR to match this on the CLI, but I'm not able to update the vscode extension to do the same.

I kept the prettify method because It's a useful abstraction that can be removed in the future if prettier is entirely removed.

PR #357, opened 2023-07-03
Add failing test for normal actions

davidkpiano:
We shouldn't use `xstate.custom` for these


farskid:
> We shouldn't use `xstate.custom` for these

Why not?

Andarist:
This is already feature-requested [here](https://github.com/statelyai/xstate-tools/issues/355) - I think it's fine to use that ticket as a tracking issue for this.

PR #359, opened 2023-07-24
Extract action properties as simple json types according to new changes to Studio serializers

farskid:


packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -160,30 +160,21 @@ describe('MachineParseResult', () => {
     expect(config.states!.a.entry).toMatchInlineSnapshot(`
       {
         "type": "xstate.custom",
```
davidkpiano, line 162: We should have some way to represent anonymous actions... not sure what it would be yet

farskid: What's wrong with representing anonymous actions as `xstate.custom`?

packages/machine-extractor/src/extractAction.ts
```diff
@@ -351,23 +268,16 @@ export const getActionCreatorName = (actionNode: ActionNode) =>
...
-      type: 'object';
-      value: Record<string, JsonEntry>;
-    } {
+): JsonExpressionString | Record<string, JsonItem> {
   if (!eventObject.properties.every((prop) => t.isObjectProperty(prop))) {
-    return {
-      type: 'expression',
-      value: fileContent.slice(eventObject.start!, eventObject.end!),
-    };
+    return `{{${fileContent.slice(eventObject.start!, eventObject.end!)}}}`;
```
Andarist, line 277: I'd introduce some util, like `asExpressionString` or smth. It's hard to tell here if we have the right amount of curly braces at the end since there are 3 (it's correct here, but it's an easy mistake to make)

packages/machine-extractor/src/extractAction.ts
```diff
@@ -351,23 +268,16 @@ export const getActionCreatorName = (actionNode: ActionNode) =>
     ? actionNode.node.callee.name
     : undefined;
 
+// TODO: enforce return type to ActionEvent so `type` property is required
```
Andarist, line 271: what should happen if somebody writes this?
```ts
raise({ notAType: 'foo' })
```

packages/machine-extractor/src/extractAction.ts
```diff
@@ -293,23 +231,18 @@ export function extractSendToAction(
             t.isObjectProperty(prop) && t.isIdentifier(prop.key),
         )
         .forEach((prop) => {
+          const name = prop.key.name;
           if (t.isStringLiteral(prop.value) || t.isBigIntLiteral(prop.value)) {
-            options[prop.key.name] = {
-              type: 'string',
-              value: prop.value.value,
-            };
+            options[name as 'id' | 'delay'] = prop.value.value;
```
Andarist, line 236: We don't have the luxury of trusting this input like that. We should filter out properties that are outside of this handled set

packages/machine-extractor/src/extractAction.ts
```diff
@@ -189,56 +152,43 @@ export function extractLogAction(
...
-  let eventObject: SendToEventArg = undefined;
-  let actorRef: SendToActorRefArg = undefined;
-  const options: { [key: string]: SendToOptionArg } = {
-    delay: undefined,
-    id: undefined,
+  let eventObject: SendToEventArg = {};
+  let actorRef: SendToActorRefArg = '';
+  const options: { id: SendToIdArg; delay: SendToDelayArg } = {
+    id: '',
+    delay: 0,
```
Andarist, line 191: This shouldn't use any defaults. What is not provided in the source code should stay as not provided, especially since there is a difference between `delay: 0` and `delay: undefined`. The first one instructs the runtime to use `setTimeout` whereas the second one doesn't (the event gets sent immediately/synchronously)

packages/machine-extractor/src/extractAction.ts
```diff
@@ -3,16 +3,17 @@ import { ActionNode } from './actions';
...
-  | { type: 'expression'; value: string };
-export type JsonValue = Pick<JsonEntry, 'value'>;
-type JsonType = Pick<JsonEntry, 'type'>;
+export type JsonItem =
+  | string
+  | number
+  | bigint
+  | boolean
+  | null
+  | undefined
```
Andarist, line 12: it's confusing to call this whole type (and related types) JSON since `undefined` is not a valid JSON value :P

farskid:
Closing this as it's superseded by https://github.com/statelyai/xstate-tools/pull/360

packages/machine-extractor/src/extractAction.ts
farskid: Thanks for catching this. I removed those defaults in the other PR.

packages/machine-extractor/src/extractAction.ts
farskid: We still need to think about proper handling of invalid inputs but for now, maybe we can trust there's a type property always in the type level?

PR #360, opened 2023-08-10
Change action structure and simplify extracted values

farskid:
This PR supersedes https://github.com/statelyai/xstate-tools/pull/359 and:

- Drops `{type, value}` structure in favor of simpler values
- Changes action structure to accept `kind` and push properties under `action` object
- Extracts expressions wrapped in doubly curly braces`{{}}`
- Extracts array and object values recursively (previously extracted as expressions)
- Drops using XState types in favor of adding local types for Action, StateNodeConfig, etc
- Makes `id` and `delay` in `sendTo` actions, optional and doesn't include them in extracted action when not present
- Creates specific parser for `SendTo` actions to treat them as builtin actions
- Updates tests accordingly


packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -158,33 +158,208 @@ describe('MachineParseResult', () => {
...
+            "expr": "{{someVar}}",
+          },
+          "kind": "inline",
         },
-      }
+      ]
+    `);
+  });
+
+  it('should extract new action structure', () => {
```
farskid, line 194: Will be removed. Just a WIP test.

packages/machine-extractor/src/__tests__/parseResult.test.ts
```diff
@@ -138,57 +138,79 @@ describe('MachineParseResult', () => {
...
+        }
+      },
+    });
+  `,
+    );
+    expect(config.states!.a.entry).toMatchInlineSnapshot(`
+      [
+        {
+          "action": {
+            "expr": "{{{type: someIdentifier, params: {foo: 'bar'}}}}",
```
davidkpiano, line 195: Is this just because we can't/don't extract the actual object yet? This should ideally be a normal action object inside of `action`, not an expression

davidkpiano: Oh I see, it's because of `someIdentifier`, sorry!

packages/machine-extractor/src/extractAction.ts
```diff
@@ -411,10 +346,23 @@ function extractEventObject(
       }
     }
   });
+  return extracted;
+}
+
+// TODO: enforce return type to ActionEvent so `type` property is required
```
farskid, line 352: @Andarist We don't yet have a validation in place, do we?  So I think this only enforces having `type` property in the type-level and is fine?

packages/machine-extractor/src/extractAction.ts
```diff
@@ -279,37 +221,32 @@ export function extractSendToAction(
...
     // Options
     if (t.isObjectExpression(arg3)) {
       arg3.properties
         .filter(
-          (prop): prop is ObjectProperyWithIdentifierKey =>
-            t.isObjectProperty(prop) && t.isIdentifier(prop.key),
+          (prop): prop is SendToOptionObjectProperty =>
+            t.isObjectProperty(prop) &&
+            t.isIdentifier(prop.key) &&
+            ['id', 'delay'].includes(prop.key.name),
```
farskid, line 236: This is now filtered out according to your comment.

packages/machine-extractor/src/__tests__/actions.test.ts
```diff
@@ -0,0 +1,1178 @@
...
+function getTestMachineConfig(configStr: string) {
+  const result = extractMachinesFromFile(configStr);
+  const machine = result!.machines[0];
+  const config = machine?.toConfig({ serializeInlineActions: true })!;
+
+  return config;
+}
+
+describe('extract actions', () => {
+  it('Ensures actions are extracted with the correct `kind`', () => {
```
Andarist, line 12: Overall I'd prefer to have more granular tests. I feel like it comes important when debugging things - it's easier to debug a single small test than to debug a noisy one when one of the tested cases in it doesn't work alright.

Not a blocker rn, just want to encourage a different style of writing test cases than this "catch-all" style

packages/machine-extractor/src/__tests__/actions.test.ts
```diff
@@ -0,0 +1,1178 @@
...
+        "action": {
+          "id": "actor",
+          "type": "xstate.stop",
+        },
+        "kind": "builtin",
+      }
+    `);
+  });
+
+  it('Should extract stop with a number id', () => {
```
Andarist, line 1154:
```suggestion
  it('Should extract stop with a bigint id', () => {
```

Andarist: although, if it's "big enough" it probably wouldn't work at runtime in XState but that's a separate issue~

packages/machine-extractor/src/actions.ts
```diff
@@ -99,6 +99,52 @@ export const ActionAsString = maybeTsAsExpression(
...
+          t.isIdentifier(prop.key) &&
+          prop.key.name === 'type' &&
+          t.isStringLiteral(prop.value) &&
+          !SUPPORTED_BUILTIN_ACTIONS.includes(prop.value.value)
+        ) {
+          return {
+            action: id,
+            node,
+            name: prop.value.value,
+            declarationType: 'object',
```
Andarist, line 132: imho this could just be `declarationType: 'named'` and if `node` turns out to be an object expression then its properties should get extracted~ in `getActionConfig`

packages/machine-extractor/src/extractAction.ts
```diff
@@ -377,31 +321,27 @@ function extractEventObject(
...
+            prop.value,
+            fileContent,
+          );
+        } else if (t.isObjectExpression(prop.value)) {
+          extracted[prop.key.name] = extractObjectRecursively(
+            prop.value,
+            fileContent,
+          );
         } else {
           console.warn(
```
Andarist, line 346: This is IMHO a little bit problematic because it favors partial results. A better approach would be to extract everything that contains unsupported things as inline. I understand though that you might not want to change this rn.

packages/machine-extractor/src/actions.ts
```diff
@@ -99,6 +99,52 @@ export const ActionAsString = maybeTsAsExpression(
...
+    const id = context.getNodeHash(node);
+    for (const prop of node.properties) {
+      if (t.isObjectProperty(prop)) {
+        if (
+          t.isIdentifier(prop.key) &&
+          prop.key.name === 'type' &&
+          t.isStringLiteral(prop.value) &&
+          !SUPPORTED_BUILTIN_ACTIONS.includes(prop.value.value)
+        ) {
+          return {
```
Andarist, line 128: those return values here already can contain `.value` - any reason why `extractEventObject` gets called from `getActionConfig` and not here?

packages/machine-extractor/src/actions.ts
```diff
@@ -213,7 +259,75 @@ export const AssignAction = wrapParserResult(
     return {
       node: result.node,
       action: assign(result.argument1Result?.value || defaultAction),
-      name: '',
+      name: 'assign',
```
Andarist, line 262: thinking out loud: we could also hoist `.kind` to this return value here - so we could leverage exhaustive switches when consuming all `.kind: 'builtin'`+`.name` in `getActionConfig`. Right now we rely on a `default:` fallback that extracts unmatched things as inline

packages/machine-extractor/src/extractAction.ts
```diff
@@ -351,24 +272,47 @@ export const getActionCreatorName = (actionNode: ActionNode) =>
...
+        }
+      }
+    }
+  });
+  return extracted;
+}
 
-  const extracted: Record<string, JsonEntry> = {};
-  eventObject.properties.forEach((prop) => {
+export function extractObjectRecursively(
```
Andarist, line 310: There is a good chunk of duplication between `extractObjectRecursively` (btw. that name implies that there is some non-recursive variant somewhere and I don't think that's the case) and `extractArrayRecursively` (and some other places, such as what happens for `assign`).

You could have `extractValue` that would handle all of those cases in the same way (for `assign` u'd call it separately for each property value that u don't want to extract as an expression)

packages/machine-extractor/src/extractAction.ts
```diff
@@ -279,37 +221,32 @@ export function extractSendToAction(
...
-              type: 'string',
-              value: prop.value.value,
-            };
+            options[name] = prop.value.value;
           } else if (t.isNumericLiteral(prop.value)) {
-            options[prop.key.name] = {
-              type: 'number',
-              value: prop.value.value,
-            };
+            options[name as 'delay'] = prop.value.value;
```
Andarist, line 243: a surprising cast here ;p what should happen with `id: 42`?

packages/machine-extractor/src/extractAction.ts
Andarist: hm, ye - the whole thing is kinda the "best effort" thing and there is no much validation in the whole thing. I think though that this shouldn't exactly be validated here.

We **want** to handle invalid things. I can edit (and fix) `sendTo({})` in the IDE and I should do the same with it in the Studio. So the goal is to transfer such a thing successfully to the Studio and validate it there + offer a way to correct the mistake

packages/machine-extractor/src/types.ts
```diff
@@ -38,4 +78,64 @@ export interface FileExtractResult<
...
+export type ExtractorAssignAction = BaseBuiltinAction<{
+  assignment:
+    | Record<string, JsonItem | JsonExpressionString>
+    | JsonExpressionString;
+}>;
+export type ExtractorRaiseAction = BaseBuiltinAction<{
+  event: Record<string, JsonItem> | JsonExpressionString;
+}>;
+export type ExtractedLogAction = BaseBuiltinAction<{
+  expr: string | JsonExpressionString;
```
Andarist, line 101: this is equivalent of `string` - so the q is: what value does it bring us here? it has some documentation value of the intent but one that's only visible in this file, if we hover over those properties in different places we won't see it

packages/machine-extractor/src/types.ts
```diff
@@ -38,4 +78,64 @@ export interface FileExtractResult<
...
+  | ExtractorStopAction;
+
+export type MachineAction = NamedAction | InlineAction | BuiltinAction;
+
+// These types are copied over from studio blocks.
+// array and object are extracted as expressions so Studio render them correctly when exporting
+export type JsonItem =
+  | string
+  | number
+  | bigint
```
Andarist, line 135: bigint is not a valid JSON value - so this type should be somehow renamed, maybe `TransferrableItem`?

packages/machine-extractor/src/extractAction.ts
farskid: What key value pair would not be extracted here? I want to understand the partial part.

packages/machine-extractor/src/actions.ts
farskid: Good idea. Done.

packages/machine-extractor/src/types.ts
farskid: Just educational. I'd say when we later look at the source of these type definitions, it's valuable to still see if a string can be an expression string too or not.

packages/machine-extractor/src/actions.ts
farskid: `extractObjectRecursively` is called inside `getActionConfig` because we need to access fileContent to slice up values that can't be transferred other than as a string like a regular expression value, etc.

packages/machine-extractor/src/actions.ts
farskid: The check to determine whether it's an object that needs to be specially treated is more complicated than that. It checks if it's an object that has a `type` property that has a string value and the value is not a built-in XState action name that we support in the Studio UI.

PR #369, opened 2023-09-19
`internal` transition extracting

farskid:


packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -337,6 +337,10 @@ export const getTransitions = (
     if (transition?.description) {
       toPush.description = transition?.description.value;
     }
+    // XState transitions are external by default so only add `internal` if its true
```
Andarist, line 340: `target: '.child'` (without the explicit `internal` prop) will still be handled correctly further down the road, right?

.changeset/brave-readers-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/machine-extractor': minor
+---
+
+Parse `internal` property on transitions
```
Andarist, line 5:
```suggestion
Extract `internal` property on transitions
```

packages/machine-extractor/src/toMachineConfig.ts
farskid: Can you give me an example and what's expected?

Andarist: The Studio should correctly set the external prop on its side to `external: false` for a transition defined like this `target: '.child'`. 

This also means that you should **always** extract `internal` property with boolean values and not only when it's set to `internal: true`. That is because this should be displayed as an external transition by the Studio:
```ts
target: '.child',
internal: false
```

farskid: I haven't checked but I do believe if `internal` is missing, the Studio assumes it's an external transition.

Andarist: If the target's descriptor "kind" is ignored then this would be a pretty important bug to fix on the Studio's side. To fix it we'll definitely need to extract both `internal: false` and `internal: true` here.

farskid: Let's stay on the safe side and always include the internal property.

packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -337,6 +337,10 @@ export const getTransitions = (
...
+    if (transition.internal?.value === true) {
```
Andarist, line 341:
```suggestion
    if (typeof transition.internal?.value === 'boolean') {
```

packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -337,6 +337,10 @@ export const getTransitions = (
...
+      toPush.internal = true;
```
Andarist, line 342:
```suggestion
      toPush.internal = transition.internal.value;
```

Andarist (approved):

PR #370, opened 2023-09-19
Extract machine context

farskid:


davidkpiano (approved):

.changeset/calm-balloons-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/machine-extractor': minor
+---
+
+Parse machine context into an object
```
Andarist, line 5:
```suggestion
Extract machine context into an object
```

packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -33,6 +33,7 @@ export interface ToMachineConfigOptions {
 const parseStateNode = (
   astResult: StateNodeReturn,
   opts: ToMachineConfigOptions | undefined,
+  isRootNode: boolean,
```
Andarist, line 36: instead of adding a parameter here, I'd introduce a wrapper function that would handle `context` and then call a recursive function for handling states. This way we just introduce a more specialized handler at the top but the core handler (this one here) stays ignorant to the `isRootNode`

packages/machine-extractor/src/toMachineConfig.ts
```diff
@@ -48,6 +49,13 @@ const parseStateNode = (
     config.type = astResult.type.value as any;
   }
 
+  if (isRootNode && t.isObjectExpression(astResult.context?.node)) {
```
Andarist, line 52: what would happen with `context: () => ({})`? I mean, not here - but further down the road

packages/machine-extractor/src/toMachineConfig.ts
farskid: Check it out again.

packages/machine-extractor/src/toMachineConfig.ts
farskid: We don't support lazy context in Studio yet so that's ignored for now.

Andarist: Are there any downsides to that? What if we export the code after some changes? Would the lazy context be completely lost in such a scenario?

farskid: Extracting lazy context doesn't benefit us other than not being lossy to that data. It won't be editable, won't be rendered in context panel and won't certainly be integrated with context schema. I'm not against supporting it if it's just a read-only context expression.

Andarist: >  I'm not against supporting it if it's just a read-only context expression.

Yeah, I was thinking about something like that for now.

farskid: Done.

Andarist (approved):

packages/machine-extractor/src/__tests__/context.test.ts
```diff
@@ -0,0 +1,91 @@
+import { extractMachinesFromFile } from '..';
+
+describe('Context parsing', () => {
```
Andarist, line 3:
```suggestion
describe('Context', () => {
```

packages/machine-extractor/src/__tests__/context.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  it('Should be able to parse machine context into an object', () => {
```
Andarist, line 4:
```suggestion
  it('Should be able to extract machine context into an object', () => {
```

farskid:
@Andarist Should we bump the version too?

Andarist:
What do u mean? this PR already has a changeset 🤔

farskid:
> What do u mean? this PR already has a changeset 🤔 

Ag, disregard that comment. I was confused for a moment.

PR #372, opened 2023-09-24
Update to Prettier 3 and latest turbo repo

mellson:
Update xstate-tools to use:

* Prettier 3 (to support uniform optimize imports across editors)
* Latest turborepo (with external config file)


mellson:
It broke some tests, so I'm abandoning it for now as this is unimportant.

PR #380, opened 2023-10-09
Add `xstate sky` cmd to the CLI

mellson:
This PR adds the `xstate sky` cmd to the `@xstate/CLI` package. This is the first step towards our Stately Sky tooling. This will allow users to run `xstate sky` on their source files as we do with `typegen`.

The command will:

- Check if a Sky config has been fetched, and skip if true.
- If no config has been fetched, it will expand the short Sky URL and try to fetch a config.
- If a valid config was fetched, it will write a sky file and update the file that consumes it.

Later, after we have tested this all works, we will add the same functionality to the VS Code extension to make all this happen automatically on save.

Currently, the Sky tests are not fully correct. They run from a stale example. Sky needs XState v5, and this repo is still on v4, so it's tricky to make it work properly before it is updated to v5.


.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
```
mellson, line 14: We need 16+ to be compatible with the Sky SDK, so I bumped it to 18.

kevinmaes (approved):
This looks good and we can test it later.

Issue #387, opened 2023-10-12
@xstate/cli typegen regression from 0.5.2


laat:
When generating types with 0.5.4, type checking with typescript gives the error `'actionA' does not exist in type 'MachineOptionsActions...`

It worked in 0.5.2

```ts
import { createMachine } from 'xstate';

const machine = createMachine(
  {
    schema: {
      services: {} as {
        getData: {
          data: any;
        };
      },
    },
    tsTypes: {} as import('./foo.typegen').Typegen0,
    predictableActionArguments: true,

    initial: 'A',
    context: {
      value: '',
    },
    states: {
      A: {
        invoke: {
          src: 'getData',
          onDone: {
            actions: ['actionA', 'actionB'],
            target: 'DONE',
          },
        },
      },
      B: {
        invoke: {
          src: 'getData',
          onDone: {
            actions: ['actionA', 'actionB'],
            target: 'DONE',
          },
        },
      },
      DONE: {},
    },
  },
  {
    actions: {
      actionA: () => {}, // <-- ts error "'actionA' does not exist in type ..."
      actionB: () => {},
    },
  }
);
```

The only difference from the typegen.ts file generated in 0.5.2 is this:
```diff
   eventsCausingActions: {
-    actionA:
-      | 'done.invoke.(machine).A:invocation[0]'
-      | 'done.invoke.(machine).B:invocation[0]';
     actionB:
       | 'done.invoke.(machine).A:invocation[0]'
       | 'done.invoke.(machine).B:invocation[0]';
   };
```

laat:
thanks! v0.5.5 solved it 👍

PR #388, opened 2023-10-12
Fixed an issue with `invoke.onDone`/`invoke.onError` actions being incorrectly ignored

Andarist:
fixes https://github.com/statelyai/xstate-tools/issues/387

TODO:
- [x] add a test case

### DIFF
--- a/packages/shared/src/forEachAction.ts
+++ b/packages/shared/src/forEachAction.ts
@@ -155,11 +155,10 @@ const forEachActionRecur = (
           replaceOrDeleteActions(doneTransition, 'actions', visitor);
         }
       }
-    }
-    /**
-     * invoke.onError transitions
-     */
-    if (event.startsWith('error.invoke')) {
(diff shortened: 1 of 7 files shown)


davidkpiano (approved):
Code LGTM

packages/shared/src/forEachAction.ts
```diff
@@ -155,11 +155,10 @@ const forEachActionRecur = (
           replaceOrDeleteActions(doneTransition, 'actions', visitor);
         }
       }
-    }
-    /**
-     * invoke.onError transitions
-     */
-    if (event.startsWith('error.invoke')) {
+    } else if (event.startsWith('error.invoke')) {
```
farskid, line 158: Should this be `error.platform`?

Andarist: It kinda should but we are also not quite relying on this here anyway. A better "fix" for this would be to just manipulate the config directly anyway so I decided not to touch this for the time being.

PR #391, opened 2023-10-13
Fix problem with Prettier version conflicts for Sky configs

mellson:
This PR fixes two problems with writing Sky configs:

- It fixes a potential version conflict between Prettier 2 & 3 by not using the prettier config of the user for Sky configs.
- It fixes [an unknown type: "ChainExpression" error](https://github.com/prettier/prettier/issues/15489) by upgrading to Prettier 2.8.8

### DIFF
--- a/apps/cli/src/bin.ts
+++ b/apps/cli/src/bin.ts
@@ -57,7 +57,7 @@ program
   )
   .argument('<files>', 'The files to target, expressed as a glob pattern')
   .option('-w, --watch', 'Run sky in watch mode')
-  .option('-f, --force', 'Always overwrite existing sky configs')
+  .option('-r, --refetch', 'Always refetch and overwrite existing sky configs')
   .option(
     '-k, --api-key <key>',
     'API key to use for interacting with the Stately Studio',
@@ -66,7 +66,7 @@ program
     async (
       filesPattern: string,
       opts: {
-        force?: boolean;
+        refetch?: boolean;
         watch?: boolean;
         apiKey?: string;
         host?: string;
@@ -80,7 +80,7 @@ program
           writeConfigToFiles({
             uri,
             apiKey,
-            forceFetch: opts.force === true,
+            forceFetch: opts.refetch === true,
             writeToFiles,
             cwd,
           }).catch((e) => {
@@ -98,7 +98,7 @@ program
               writeConfigToFiles({
                 uri,
                 apiKey,
-                forceFetch: opts.force === true,
+                forceFetch: opts.refetch === true,
                 writeToFiles,
                 cwd,
               }),
--- a/apps/cli/src/bin.ts
+++ b/apps/cli/src/bin.ts
@@ -57,6 +57,7 @@ program
   )
   .argument('<files>', 'The files to target, expressed as a glob pattern')
   .option('-w, --watch', 'Run sky in watch mode')
+  .option('-f, --force', 'Always overwrite existing sky configs')
   .option(
     '-k, --api-key <key>',
     'API key to use for interacting with the Stately Studio',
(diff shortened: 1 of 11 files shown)


packages/machine-extractor/src/sky/skyConfigModifySource.ts
```diff
@@ -66,13 +66,9 @@ export const modifySkyConfigSource = async (opts: { filePath: string }) => {
     });
 
     const output = recast.print(ast).code;
-    const prettierConfig = await prettier.resolveConfig(opts.filePath);
```
mellson, line 69: This could create a version conflict with Prettier where it would use the user's Prettier config, which might not be compatible with the bundled Prettier.

Andarist: Would the problem get largely addressed if we just wouldn't bundle all of ur deps? I noticed this just yesterday and I think that we shouldn't be using `--bundle` like this here:
https://github.com/statelyai/xstate-tools/blob/0495b7489b871b10ddbcf3a31218e2d8bcdc649d/apps/cli/package.json#L11

mellson: Yes, that seems to work as well. Thanks for the feedback. I've updated the build script in the latest commit.

mellson: Wait, that didn't entirely work after a restart of my IDE. So, I'll remove the Prettier config part again but keep it unbundled.

mellson: Going back to the bundled version again. I couldn't get it to not fail with the unbundled version. So maybe the unbundling is something we can handle in another PR?

Andarist: > So maybe the unbundling is something we can handle in another PR?

Yes, let's leave it out.

> This could create a version conflict with Prettier where it would use the user's Prettier config, which might not be compatible with the bundled Prettier.

That's why we try to honor the local version of Prettier in other places where we use it:
https://github.com/statelyai/xstate-tools/blob/2c8387fa9fc9ad26817cf90a7b9d54067fa8395c/apps/cli/src/typegen/writeToFiles.ts#L26

I think that a good chunk of `modifySkyConfigSource` doesn't belong to the `machine-extractor` package. If you'd move the FS interaction to `@xstate/cli` then you'd have an easier time using Prettier correctly there.

mellson: Thanks, I moved to use the same `getPrettierInstance` function and did a mini-refactor of `modifySkyConfigSource `

apps/cli/src/bin.ts
```diff
@@ -57,21 +57,33 @@ program
   )
   .argument('<files>', 'The files to target, expressed as a glob pattern')
   .option('-w, --watch', 'Run sky in watch mode')
+  .option('-f, --force', 'Always overwrite existing sky configs')
```
Andarist, line 60: maybe `--refetch`?

packages/machine-extractor/package.json
```diff
@@ -12,7 +12,7 @@
     "@babel/parser": "^7.21.4",
     "@babel/traverse": "^7.21.4",
     "@babel/types": "^7.21.4",
-    "prettier": "^2.8.7",
+    "prettier": "^2.8.8",
```
Andarist, line 15: i believe that now machine-extractor removed its dependency on prettier so this could be removed altogether

mellson: Good catch, removed

apps/cli/src/sky/writeConfigToFiles.ts
```diff
@@ -57,8 +60,19 @@ export const writeConfigToFiles = async (opts: {
...
-            console.log(`${opts.uri} - updated with sky config`);
+            const fileContents = await fs.readFile(opts.uri, 'utf8');
+            const code = await modifySkyConfigSource({
+              fileContents,
+              filePath: opts.uri,
+            });
+            if (code) {
+              const formattedCode = await getPrettierInstance(opts.cwd).format(
+                code,
+                { parser: 'typescript' },
```
Andarist, line 71: in the other place we are doing smth like this
```suggestion
                {
                  ...(await prettierInstance.resolveConfig(opts.uri)),
                  parser: 'typescript'
                },
```

packages/machine-extractor/src/sky/skyConfigModifySource.ts
Andarist: note that this is a breaking change - but since sky is experimental right now, I'm OK with this and I think that the refactor that was made here is an improvement 👍

apps/cli/src/sky/writeConfigToFiles.ts
mellson: updated in latest commit

Andarist (approved):

PR #396, opened 2023-11-10
Extract guard implementations

farskid:


packages/machine-extractor/src/__tests__/guards.test.ts
```diff
@@ -0,0 +1,190 @@
...
+              "f1": {},
+            },
+          },
+          "type": "myGuard",
+        },
+        "kind": "named",
+      }
+    `);
+  });
+  it.only('Params should be ignored for parameterized actions with params set to any value other than an object expression', () => {
```
davidkpiano, line 146:
```suggestion
  it('Params should be ignored for parameterized actions with params set to any value other than an object expression', () => {
```

farskid:
Superseded by https://github.com/statelyai/xstate-tools/pull/42

PR #397, opened 2023-11-13
Enable inline strings and env variables to be Sky API keys

mellson:
This PR makes it possible for the `xstate sky` cmd to pickup API Keys defined in environment variables compatible with node, such as `process.env.SOME_API_KEY`


davidkpiano (approved):

packages/machine-extractor/src/identifiers.ts
```diff
@@ -150,7 +150,7 @@ export const objectExpressionWithDeepPath = <Result>(
 const getRootIdentifierOfDeepMemberExpression = (
   deepMemberExpression: DeepMemberExpression | undefined,
 ): t.Identifier | undefined => {
-  if (!deepMemberExpressionToPath) return undefined;
+  if (!deepMemberExpression) return undefined;
```
Andarist, line 153: 😅

mellson: Yeah, this one was fun. It caused a `Maximum call stack size exceeded` because it just kept checking if the function  `deepMemberExpressionToPath` was still defined 😅

packages/machine-extractor/src/sky/skyConfigExtractResult.ts
```diff
@@ -1,14 +1,19 @@
 import * as t from '@babel/types';
 import { hashedId } from '../utils';
 import { SkyConfigCallExpression } from './skyConfigCallExpression';
+import { getClosestEnvFile, getEnvValue } from './skyEnvFileUtils';
 
 export function skyConfigExtractResult({
   file,
```
Andarist, line 7: It's a little bit yucky that we have to pass 3 file-related things here. It's totally fine right now - I just with that `file` would have at least `filePath` on it (but it doesn't and it's disappointing 😢 )

packages/machine-extractor/src/sky/skyEnvFileUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+/**
+ * Returns the value of a given key in an environment variable string.
+ * @param envString - The environment variable string to search in.
+ * @param key - The key to search for.
+ * @returns The value of the key if found, otherwise undefined.
+ */
+export function getEnvValue(envString: string, key: string) {
```
Andarist, line 50: you could use the `dotenv` package and its `dotenv.parse` here

packages/shared/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "dist/xstate-tools-shared.cjs.js",
   "license": "MIT",
   "dependencies": {
-    "@statelyai/sky": "0.0.4",
+    "@statelyai/sky": "0.0.7",
```
Andarist, line 7: Does the `@xstate-tools/shared` package need to depend on `@statelyai/sky`? 🤔

packages/machine-extractor/src/sky/skyEnvFileUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as path from 'path';
+
+/**
+ * Searches for the closest '.env' file starting from the directory of the given file path and moving up the directory tree until the root of the project is reached.
+ * @param filePath - The path of the file to start searching from.
+ * @param cwd - The root directory of the project.
+ * @returns The contents of the closest '.env' file.
+ * @throws An error if no '.env' file is found in the project.
+ */
+export function getClosestEnvFile({
```
Andarist, line 11: there is a handy [`find-up`](https://github.com/sindresorhus/find-up) package that you could use here, it also supports `stopAt` option that you are implementing here implicitly

packages/machine-extractor/src/sky/skyEnvFileUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+}) {
+  const startDir = path.dirname(filePath);
+  if (!fs.existsSync(startDir)) {
+    throw new Error(`No such directory: ${startDir}`);
+  } else {
+    let currentDir = startDir;
+    let searchingInsideProject = true;
+    while (searchingInsideProject) {
+      const fileList = fs.readdirSync(currentDir);
+      const targetFile = fileList.find((file) => file.startsWith('.env'));
```
Andarist, line 26: this could be replaced with `existsSync`

packages/shared/package.json
mellson: It does not. I have removed it in the latest commit

packages/machine-extractor/src/sky/skyConfigNode.ts
```diff
@@ -18,11 +22,42 @@ export type SkyConfigNodeReturn = WithValueNodes<{
...
+      createParser({
+        babelMatcher: t.isMemberExpression,
+        parseNode: (node, context): StringLiteralNode => {
+          if (!context.getNodeSource || !context.getEnvVariable) {
+            throw new Error("Couldn't find API key in any of the env files");
+          }
+          // Let's find the last part of the expression, e.g. `API_KEY` in `process.env.API_KEY`
+          const envVariableName = context
+            .getNodeSource(node)
+            .match(/(?<=\.)(\w+)(?!.*\.)/);
```
Andarist, line 38: instead of looking into this as a string you could do smth like this:
```ts
const envVariableName = t.buildMatchMemberExpression('process.env')(node.object) && t.isIdentifier(path.node.property) ? node.property.name : null
```

packages/machine-extractor/src/sky/skyConfigNode.ts
```diff
@@ -18,11 +22,42 @@ export type SkyConfigNodeReturn = WithValueNodes<{
 }> &
   Pick<ObjectPropertyInfo, 'node'>;
 
+const StringLiteralOrExpression = unionType([
```
Andarist, line 25:
```suggestion
const StringLiteralOrProcessEnvKey = unionType([
```

mellson: I updated the name to `StringLiteralOrEnvKey`. I think the `Process` part was too specific as this also works for `import.meta.SOME_KEY`

packages/machine-extractor/src/sky/skyConfigNode.ts
Andarist: Since you also want to handle `import.meta` then
```ts
const envVariableName = (t.isMetaProperty(node.object) || t.buildMatchMemberExpression('process.env')(node.object)) && t.isIdentifier(node.property)
 ? node.property.name
  : null
```

Andarist: Please recheck if that works. I tried to quickly confirm that using the https://astexplorer.net/ and I think it's fine but better to be safe than sorry

mellson: It works great; I tested it before adding it in. Using both `process.env.KEY` and `import.meta.KEY`

Andarist (approved):

packages/machine-extractor/src/sky/skyConfigExtractResult.ts
```diff
@@ -17,6 +22,13 @@ export function skyConfigExtractResult({
       const fileText = fileContent.substring(node.start!, node.end!);
       return hashedId(fileText);
     },
+    getNodeSource: (node: t.Node): string => {
+      return fileContent.substring(node.start!, node.end!);
+    },
+    getEnvVariable: (name: string): string | undefined => {
```
Andarist, line 28: one last thing here - it likely should be possible to read the value from actual `process.env` too. This way you could override whatever can be found in `.env` or you could just use env vars if you are not using env files at all.

So I'd propose smth like this here, at the top of the function:
```ts
if (process.env[name]) {
  return process.env[name];
}
```

PR #405, opened 2023-11-16
Machine options and support v5 `guard` and `actors`

farskid:
- Support both `actors` and `guards` in machine implementations and if both are present, merges them into a single object
- Support v5 `guard` keyword for transition conditions and choose action conditions. If both `cond` and `guard` are present on a transaction, picks `guard`.


packages/machine-extractor/src/__tests__/options.test.ts
```diff
@@ -21,4 +21,58 @@ describe('Options', () => {
 
     expect(t.isObjectMethod(node)).toBeTruthy();
   });
+
+  it.only('Should extract implementations provided in machine options', () => {
```
davidkpiano, line 25:
```suggestion
  it('Should extract implementations provided in machine options', () => {
```

packages/machine-extractor/src/extractMachinesFromFile.ts
```diff
@@ -13,15 +13,17 @@ export const extractMachinesFromFile = (
...
   return {
-    machines: machineNodes.map((node) =>
-      getMachineExtractResult({
-        file,
-        fileContent,
-        node,
-      }),
-    ),
+    machines,
+    options: machines.map((m) => m?.getAllMachineOptions()),
```
Andarist, line 26: Why do we need this here? Can't we just always get *specific* options from a *specific* machine when needed?

packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -708,6 +709,46 @@ export class MachineExtractResult {
     return delays.toObject();
   };
 
+  getAllMachineOptions = () => {
```
Andarist, line 712: we are not really super strict about it today but I think that currently we call those `implementations` (and not `options`)

davidkpiano (approved):

Andarist:
> Support both actors and guards in machine implementations and if both are present, merges them into a single object

This sounds incorrect. `guards` and `actors` probably shouldn't be merged into a single object. Or have you maybe meant `actors` and `services`?

.changeset/fair-steaks-kneel.md
```diff
@@ -0,0 +1,7 @@
+---
+'@xstate/machine-extractor': minor
+---
+
+- Support both actors and guards in machine implementations and if both are present, merges them into a single object
```
Andarist, line 5: This sounds incorrect. `guards` and `actors` probably shouldn't be merged into a single object. Or have you maybe meant `actors` and `services`?

.changeset/fair-steaks-kneel.md
Andarist, line 1: This changeset has 2 unrelated things listed here - those should just be 2 separate changesets

packages/machine-extractor/src/MachineExtractResult.ts
```diff
@@ -708,6 +709,46 @@ export class MachineExtractResult {
...
+    ) {
+      return { actions: {}, actors: {}, guards: {}, delays: {} };
+    }
+
+    const out: Record<
+      'actions' | 'actors' | 'guards' | 'delays',
+      Record<string, string>
+    > = { actions: {}, actors: {}, guards: {}, delays: {} } as const;
+
+    const foundKeyToOutKey: Record<string, keyof typeof out> = {
```
Andarist, line 725: nit: IMHO this is a little bit overcomplicated for what it is. A much simpler version would be something like:
```ts
return {
  actors: { ...getThing(this.machineCallResult.options.services), ...getThing(this.machineCallResult.options.actors) },
  actions: getThing(this.machineCallResult.options.actions)
  delays: getThing(this.machineCallResult.options.delays),
  guards: getThing(this.machineCallResult.options.guards)
}
```

.changeset/fair-steaks-kneel.md
farskid: Yeah, it's a typo. It's supposed to be `actors` and `services`.

packages/machine-extractor/src/MachineExtractResult.ts
farskid: Sure, I'll update.

.changeset/fair-steaks-kneel.md
farskid: So 2 separate changesets for a single PR or two separate PRs?

Andarist: 2 separate changesets work OK

.changeset/fair-steaks-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xstate/machine-extractor': minor
+---
+
+Support both actors and guards in machine implementations and if both are present, merges them into a single object
```
Andarist:
```suggestion
Support both `actors` and `services` in machine implementations and if both are present, merges them into a single object
```

Andarist (approved):

PR #417, opened 2023-12-20
Do not stringify null literals as object values

farskid:


packages/machine-extractor/src/extractAction.ts
```diff
@@ -330,11 +330,13 @@ export function extractObjectRecursively(
 
   object.properties.forEach((prop) => {
     if (t.isObjectProperty(prop)) {
+      const key = getObjectPropertyKey(prop);
```
Andarist, line 333: It looks like it might be unused
```suggestion
```

Andarist (approved):

.vscode/settings.json
```diff
@@ -2,8 +2,8 @@
   "editor.insertSpaces": false,
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.sortImports": true
+    "source.fixAll.eslint": "explicit",
+    "source.sortImports": "explicit"
```
farskid, line 6: Freaking VSCode.

PR #420, opened 2023-12-21
Basic v5 support in machine extractor

Andarist:


packages/machine-extractor/src/__tests__/modifications/edit_guard_v5.test.ts
```diff
@@ -0,0 +1,82 @@
+import outdent from 'outdent';
```
with-heart, line 1: Premium test content 😂

with-heart: The whole file, not just this line (thanks github mobile)

davidkpiano (approved):

apps/extension/client/src/initiateEditor.ts
```diff
@@ -27,6 +27,7 @@ type OpenLinkEvent = {
 type MachineChangedEvent = {
   type: 'MACHINE_CHANGED';
   edits: MachineEdit[];
+  options?: { v5?: boolean } | undefined;
```
mellson, line 30:
```suggestion
  options?: { xstateVersion?: 4 | 5 } | undefined;
```
Should we make it more explicit which version we want? This will also set us up for future releases.

mellson (approved):
Looks good, thanks for doing this so quickly 👍 Up to you if you want to change the `ModifyOptions` to be more explicit about which version it's using.

farskid (approved):
Code looks good to me.

apps/extension/client/src/initiateEditor.ts
Andarist: This will likely change in the future anyway so I'm not as concerned about it. It's also within a semi-internal package so we can always bump a major version if we need to change this

PR #422, opened 2023-12-29
Extract inline actors

farskid:
Todo:
- [ ] tests

### DIFF
--- a/packages/machine-extractor/src/extractAction.ts
+++ b/packages/machine-extractor/src/extractAction.ts
@@ -270,9 +270,9 @@ export function extractSendToAction(
 // Todo: support namespace imports and aliased specifiers
 // import * as actions from 'xstate'; actions.sendTo
 // import {sendTo as x} from 'xstate'
-export const getActionCreatorName = (actionNode: ActionNode) =>
-  t.isCallExpression(actionNode.node) && t.isIdentifier(actionNode.node.callee)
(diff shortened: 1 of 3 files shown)


apps/cli/src/sky/writeToFetchedMachineFile.ts
```diff
@@ -32,7 +32,7 @@ export const writeSkyConfig = async (opts: {
   }
   const prettierConfig = await prettier.resolveConfig(filePath);
   const pathToSave = getPathToSave(filePath);
-  const code = removeLeadingExport(skyConfig.configString);
+  const code = removeLeadingExport(skyConfig.prettyConfigString);
```
farskid, line 35: @mellson check if this is correct?

mellson: It is not, it should remain `configString`

farskid: Thy property doesn't exit on the sky config type.

mellson: Sounds like you might be on an older version of the `@statelyai/sky` package. Try running `yarn i` at the root of the repo, this should install a version where `configString` is there.

davidkpiano (approved):

PR #424, opened 2024-01-04
Extract inline guards

farskid:
Guards are now extracted with the `kind` property as either named or inline.

TODO:
- [x] tests


davidkpiano (approved):
Tests pass, LGTM

PR #428, opened 2024-01-10
Basic v5 support in the extension

Andarist:


farskid (approved):
The code looks good to me. I'm not sure if there's a way to test it so I approve.

apps/extension/server/src/server.ts
```diff
@@ -202,8 +208,105 @@ connection.onCodeLens(({ textDocument }) => {
...
   deepEqual(typegenData.typesNode.value, typegenData.data.tsTypesValue);
 
+// this shouldn't use file system directly but rather an injected module
+// atm this extension doesn't work in browsers anyway though
+async function isXState5(uri: string) {
+  try {
+    const pkgJsonPath = require.resolve('xstate/package.json', {
+      paths: [uriToFileName(UriUtils.dirname(URI.parse(uri)).toString())],
+    });
+    return (require(pkgJsonPath).version as string).startsWith('5');
```
davidkpiano, line 218: What if it's like `"^5.4.1"`?

Andarist: It can’t be. I’m checking XState’s package.json and its version here - not a dependency range used to install XState in the project

davidkpiano: Ahh okay.

PR #432, opened 2024-01-11
Extract state config

farskid:
Restructure, extract and test:
- `state.states`
- `state.initial`
- `state.type`
- `state.history`


farskid:
@Andarist Lint is failing due to an unchanged file in this PR.

PR #435, opened 2024-01-11
Extract base state config

farskid:
Restructure, extract and test:
- `state.states`
- `state.initial`
- `state.type`
- `state.history`


.changeset/quiet-melons-train.md
Andarist, line 1: We don't care about changesets at the current stage of development, let's remove this and avoid adding changesets in the future PRs until we start publishing things.

new-packages/ts-project/__tests__/createMachine.test.ts
```diff
@@ -16,23 +16,24 @@ test('should extract empty config', async () => {
...
+          "states": [],
+        },
         [],
       ],
     ]
   `);
 });
 
-test('should extract root state keys', async () => {
+test('should extract createMachine empty config', async () => {
```
Andarist, line 30:
```suggestion
test('should extract createMachine with empty config', async () => {
```

new-packages/ts-project/__tests__/createMachine.test.ts
Andarist, line 1: TODO: add a test case that shows that multiple machines can be extracted per file

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -0,0 +1,695 @@
+import { expect, test } from 'vitest';
+import { createTestProject, testdir, ts } from './utils';
+
+test('should extract states recursively', async () => {
```
Andarist, line 4: imho a test file shouldn't start with something as "complex" as a "should extract states recursively". We need simpler tests first 😉

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -0,0 +1,695 @@
...
+          },
+          "states": [],
+        },
+        [],
+      ],
+    ]
+  `);
+});
+
+test('Should extract state.initial with string value', async () => {
```
Andarist, line 119: nit: the inconsistency with the title titles starting with either uppercase or a lowercase kills me. Personally, I'd prefer sticking to a lowercase.

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -0,0 +1,695 @@
...
+              "type": undefined,
+            },
+            "id": "",
+            "states": [],
+          },
+          "states": [],
+        },
+        [
+          {
+            "node": NodeObject {
```
Andarist, line 262: We should remove `node` from errors in a followup PR. It's super noisy and we might not want to use this directly anyhow. I think it's better to attach custom metadata / text spans to those error objects. We don't have to spend time on this at all right now though. It's enough to recognize the errors for now and we'll take care of being able to actually display them later on.

new-packages/ts-project/src/state.ts
```diff
@@ -0,0 +1,203 @@
+import type { Expression } from 'typescript';
+import { ExtractionContext, ExtractorStateConfig } from './types';
+import { getPropertyKey } from './utils';
+
+const hasUnefinedValue = (ts: typeof import('typescript'), prop: Expression) =>
```
Andarist, line 5:
```suggestion
const isUndefined = (ts: typeof import('typescript'), prop: Expression) =>
```

.changeset/quiet-melons-train.md
farskid: That's right. I only added cause it higher we might be using the main CI rules.

new-packages/ts-project/src/state.ts
```diff
@@ -0,0 +1,203 @@
...
+      node: state,
+    });
+    return;
+  }
+
+  for (const prop of state.properties) {
+    if (ts.isPropertyAssignment(prop)) {
+      const key = getPropertyKey(ctx, ts, prop);
+
+      /* state.states */
```
Andarist, line 53:
```suggestion
```

new-packages/ts-project/src/index.ts
```diff
@@ -183,9 +55,13 @@ export function createProject(
       if (!sourceFile) {
         return [];
       }
-      return findCreateMachineCalls(ts, sourceFile).map((call) =>
-        extractMachineConfig(ts, call),
-      );
+      return findCreateMachineCalls(ts, sourceFile).map((call) => {
+        const ctx: ExtractionContext = {
+          errors: [],
+          sourceFile,
```
Andarist, line 61: it's not apparent why this has been added here since it stays unused

new-packages/ts-project/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+      type: 'property_key_no_roundtrip';
+      node: unknown;
+    }
+  | {
+      type: 'property_key_unhandled';
+      propertyKind: 'computed' | 'private';
+      node: unknown;
+    };
+
+export type ExtractorStateConfig = {
```
Andarist, line 28: I think that we should aim for compatibility with the `DigraphDef` and not `DigraphData`. To make it happen we might have to make the digraph's initialization more flexible to handle this but there is a chance that we'll be able to handle patch-based diffs way easier. If we keep using `DigraphDef` here then we won't make anything easier - apart from the initialization of the digraph. We also want to "seed" unique ids with the ones generated here but `DigraphData` doesn't even have them (`DigraphDef` has)

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -0,0 +1,695 @@
...
+          },
+          "states": [],
+        },
+        [],
+      ],
+    ]
+  `);
+});
+
+test('Should extract state.initial with undefined value', async () => {
```
Andarist, line 180: I'm not sure if this is a good (future-proof) title. `initial: undefined` is so meaningless for us that I'd treat this as an error. One from which we might be able to recover but still an error.

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -0,0 +1,695 @@
...
+            "states": [],
+          },
+          "states": [],
+        },
+        [],
+      ],
+    ]
+  `);
+});
+test('Should extract state.type with value undefined', async () => {
```
Andarist, line 439: same as https://github.com/statelyai/xstate-tools/pull/435/files#r1449215229

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -0,0 +1,695 @@
...
+          },
+          "states": [],
+        },
+        [],
+      ],
+    ]
+  `);
+});
+
+test('Should extract state.history with value undefined', async () => {
```
Andarist, line 654: this one actually has some meaning for us, it's likely a redundant version of `history: 'shallow'`

new-packages/ts-project/src/state.ts
```diff
@@ -0,0 +1,203 @@
...
+      if (key === 'states' && ts.isObjectLiteralExpression(prop.initializer)) {
```
Andarist, line 54: This can easily be flattened~. We should be iterating over property assignments, try to get their keys and then switch/case through all known keys. This would allow you to drop `isStateNodeConfigProperty` and convert this if/elseif into a single switch/case

new-packages/ts-project/src/state.ts
```diff
@@ -0,0 +1,203 @@
...
+            });
+            continue;
+          }
+
+          state satisfies never;
+        }
+      } else if (key && isStateNodeConfigProperty(key)) {
+        switch (key) {
+          case 'initial': {
+            if (ts.isStringLiteral(prop.initializer)) {
```
Andarist, line 102: we should handle `NoSubstitutionTemplateLiteral`

new-packages/ts-project/src/state.ts
```diff
@@ -0,0 +1,203 @@
...
+            break;
+          }
+          case 'type': {
+            const isStringWithValidValue =
+              ts.isStringLiteral(prop.initializer) &&
+              ['history', 'atomic', 'compound', 'parallel', 'final'].includes(
+                prop.initializer.text,
+              );
+
+            if (isStringWithValidValue) {
```
Andarist, line 121: More nesting would be preferred here, imho. First recognize the AST node - then handle its valid values. If you don't want to have specific errors in this branch of the code then `continue` when you managed to handle a thing successfully and otherwise just let it go to the unconditional `ctx.errors.push` below if/elseif

new-packages/ts-project/src/state.ts
```diff
@@ -0,0 +1,203 @@
...
+    data: {
+      initial: result.data.initial ?? undefined,
+      type: result.data.type ?? undefined,
+      history: result.data.history ?? undefined,
+      metaEntries: result.data.metaEntries ?? [],
+      entry: result.data.entry ?? [],
+      exit: result.data.exit ?? [],
+      invoke: result.data.invoke ?? [],
+      tags: result.data.tags ?? [],
+      description: result.data.description ?? undefined,
```
Andarist, line 199: I'm not a fan of this `data` nesting 😅 This is how digraph keeps it though so we have to keep it here too :(

new-packages/ts-project/src/index.ts
farskid: Was added for a code that's already deleted. I'll delete it.

Andarist (approved):

PR #438, opened 2024-01-15
`state.history` with boolean value

farskid:


new-packages/ts-project/src/state.ts
```diff
@@ -5,6 +5,11 @@ import { getPropertyKey, uniqueId } from './utils';
 const isUndefined = (ts: typeof import('typescript'), prop: Expression) =>
   ts.isIdentifier(prop) && ts.idText(prop) === 'undefined';
 
+const isTrue = (ts: typeof import('typescript'), prop: Expression) =>
+  prop.kind === ts.SyntaxKind.TrueKeyword;
+const isFalse = (ts: typeof import('typescript'), prop: Expression) =>
+  prop.kind === ts.SyntaxKind.FalseKeyword;
```
Andarist, line 11: IMHO such short utils are somewhat redundant - the `isUndefined` makes sense because it's not that obvious how to check that but true/false are just checking the `.kind`. The advantage of not using such utils is also that type narrowing works out of the box - without any extra effort. The control flow analysis can't look into your `isTrue`/`isFalse` - to make them work you'd have to annotate those as type predicates which feels like a chore here (and inflates a simple one-liner just a tad too much)

farskid: That's a good point. I'll revert that change in https://github.com/statelyai/xstate-tools/pull/439

PR #439, opened 2024-01-15
Extract `state.meta`

farskid:


new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -1173,3 +1279,450 @@ test('should raise error when state.description has invalid value', async () =>
...
+            },
+            "root": "state-0",
+          },
+          [],
+        ],
+      ]
+    `);
+});
+
+test('should extract state.meta with undefined value', async () => {
```
Andarist, line 1579: this will require some extra thought since a property with an undefined value will get lost when transferring it from the VS Code extension to the Studio

Andarist: On top of that, undefined is stringified~ by the Studio (when I put that into the meta field)

farskid: Studio doesn't support undefined value for meta properties. They're JSON values, string, number, boolean, array, object or null. But the whole meta can be undefined.

Andarist: actually, this test is wrong/has a wrong title/or something. An undefined value is not extracted according to the snapshot and the source code

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -1173,3 +1279,450 @@ test('should raise error when state.description has invalid value', async () =>
...
+            },
+            "root": "state-0",
+          },
+          [],
+        ],
+      ]
+    `);
+});
+
+// TODO: this needs to change as technically any value is accepted by XState. Studio only supports plain js objects right now
```
Andarist, line 1652: We could make this a requirement on the XState side of things - or for the integration with the Studio. While currently, we don't enforce those to be objects... it's probably unlikely that people use non-objects here anyway

new-packages/ts-project/__tests__/state.test.ts
farskid: > actually, this test is wrong/has a wrong title/or something. An undefined value is not extracted according to the snapshot and the source code

Then we should title it as "Do not raise error for undefined state.meta" or "ignores undefined state.meta". All the tests is supposed to do is to check the extraction succeeds without error.

new-packages/ts-project/__tests__/state.test.ts
```diff
@@ -1173,3 +1279,450 @@ test('should raise error when state.description has invalid value', async () =>
...
+    'index.ts': ts`
+      import { createMachine } from "xstate";
+
+      createMachine({
+        initial: 'foo'
+        states: {
+          foo: {
+            meta: {
+              obj: {
+                foo: {
```
Andarist, line 1499: nit: i'd rename this one - `foo` appears now as a state name and as a key in the meta

new-packages/ts-project/src/state.ts
```diff
@@ -157,6 +169,30 @@ export function extractState(
           });
           break;
         }
+        case 'meta': {
+          if (ts.isObjectLiteralExpression(prop.initializer)) {
+            for (const meta of prop.initializer.properties) {
```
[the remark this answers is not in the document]
farskid: This was before you do actions and check that all items in an array are valid. I can change this now to follow the same pattern.

new-packages/ts-project/src/utils.ts
```diff
@@ -47,6 +47,62 @@ export function getPropertyKey(
...
+  if (prop.kind === ts.SyntaxKind.TrueKeyword) {
+    return true;
+  }
+  if (prop.kind === ts.SyntaxKind.FalseKeyword) {
+    return false;
+  }
+  if (prop.kind === ts.SyntaxKind.NullKeyword) {
+    return null;
+  }
+  // TODO: set a strategy here. Ignore the whole array if an items can't be extracted or try to extract as much as possible?
```
farskid, line 75: Same as https://github.com/statelyai/xstate-tools/pull/439/files#r1453405250

new-packages/ts-project/src/state.ts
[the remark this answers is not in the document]
Andarist: You can leave this - I think that we'll just end up adjusting those things when we get to bidi syncing anyway so it's not super important to spend time on this right now. Our current north star is to extract things that are understood by the Studio and render the extracted definition in the Studio.

new-packages/ts-project/__tests__/state.test.ts
Andarist: > We're talking about thinking how to do this for all properties, right? This is not specific to meta.

Ye, it might get annoying everywhere - but `meta` (and `context`, etc) will be especially annoying in this regard because those are user-provided values. For everything else, we can probably assume things since we know the expected schema~ upfront.

> Then we should title it as "Do not raise error for undefined state.meta" or "ignores undefined state.meta". All the tests is supposed to do is to check the extraction succeeds without error.

fine by me

new-packages/ts-project/src/utils.ts
```diff
@@ -47,6 +47,62 @@ export function getPropertyKey(
...
+export const isUndefined = (
+  ts: typeof import('typescript'),
+  prop: Expression,
+) => ts.isIdentifier(prop) && ts.idText(prop) === 'undefined';
+
+export function getJsonValue(
+  ctx: ExtractionContext,
+  ts: typeof import('typescript'),
+  prop: Expression,
+): JsonItem | undefined {
```
Andarist, line 59: `getJsonValue` returns `JsonItem` 😉 both should use the same term

PR #440, opened 2024-01-16
Extract machine context

farskid:


new-packages/ts-project/src/utils.ts
```diff
@@ -103,6 +108,13 @@ export function getJsonValue(
     return out;
   }
 }
+
+export const getJsonObject = (
+  ctx: ExtractionContext,
+  ts: typeof import('typescript'),
+  prop: ObjectLiteralExpression,
+) => getJsonValue(ctx, ts, prop) as JsonObject;
```
Andarist: It's not exactly what I had in mind - this just pushes the cast into this utility. I pushed out the proposed change directly to the PR to showcase what I've meant

PR #441, opened 2024-01-16
Extract entry/exit actions

Andarist:


new-packages/ts-project/src/state.ts
```diff
@@ -157,6 +182,65 @@ export function extractState(
...
+              });
+            },
+          );
+
+          if (!everyDefined(blocks)) {
+            ctx.errors.push({
+              type: 'state_property_unhandled',
+            });
+            continue;
+          }
```
farskid, line 231: So for now, we'd lose all the actions in entry for example if one of them can't be extracted?

new-packages/ts-project/src/state.ts
```diff
@@ -157,6 +182,65 @@ export function extractState(
...
+
+          for (const block of blocks) {
+            node.data[key].push(block.uniqueId);
+            ctx.digraph.blocks[block.uniqueId] = block;
+            ctx.digraph.implementations.actions[block.sourceId] ??= {
```
farskid, line 236: TIL about `??=`!

new-packages/ts-project/__tests__/actions.test.ts
```diff
@@ -0,0 +1,691 @@
...
+                  "type": "callDavid",
+                },
+                "sourceId": "callDavid",
+                "uniqueId": "block-0",
+              },
+            },
+            "edges": {},
+            "implementations": {
+              "actions": {
+                "undefined": {
```
farskid, line 37: The object key here should be "callDavid".

new-packages/ts-project/src/state.ts
Andarist: Yes, that was what we agreed on yesterday in the evening. While it's not ideal, it won't require any immediate changes in the Studio.

new-packages/ts-project/__tests__/actions.test.ts
```diff
@@ -0,0 +1,691 @@
...
+                  "id": "callAnders",
+                  "name": "callAnders",
+                  "type": "action",
+                },
```
farskid, line 261: We should see two actions here because we have two action blocks with different source IDs.

new-packages/ts-project/__tests__/actions.test.ts
```diff
@@ -0,0 +1,691 @@
...
+                "uniqueId": "block-0",
+              },
+            },
+            "edges": {},
+            "implementations": {
+              "actions": {
+                "inline:action-0": {
+                  "id": "inline:action-0",
+                  "name": "inline:action-0",
+                  "type": "action",
```
farskid, line 477: The source must have a `code` property that contains the source code of the action `() => {}`

new-packages/ts-project/__tests__/actions.test.ts
Andarist: Ah, good catch - that was a bug in the test utils. I just pushed out a fix for this.

new-packages/ts-project/__tests__/actions.test.ts
Andarist: Fixed - they got "merged" because both were keyed under `"undefined"` so the last one won this fight.

new-packages/ts-project/__tests__/actions.test.ts
Andarist: How would it be used by the Studio?

farskid: If we don't show the source panel and exported code, it won't be used at all.

farskid (approved):

PR #442, opened 2024-01-18
Extract actors

farskid:


new-packages/ts-project/src/state.ts
```diff
@@ -30,6 +29,40 @@ const createActionBlock = ({
...
+    parentId,
+    sourceId,
+    properties: {
+      src: sourceId,
+      id: actorId,
+    },
+  };
+}
+
+function getActorId(
```
Andarist, line 54: overall this isn't a bad idea but I struggle to see why we are doing this here now if we didn't do this for anything else? 🤔 What makes this special?

farskid: Do you mean the function itself or what we're doing to the actor id?

new-packages/ts-project/src/state.ts
```diff
@@ -30,6 +29,40 @@ const createActionBlock = ({
...
+  ts: typeof import('typescript'),
+  parentId: string,
+  index: number,
+  idProperty: PropertyAssignment | undefined,
+) {
+  if (idProperty && ts.isStringLiteralLike(idProperty.initializer)) {
+    return idProperty.initializer.text;
+  }
+  // TODO: replace this with `:serializableId:invocation[:index]`
```
Andarist, line 63: why? even including `parentId` here is weird-ish to me. What happens when you reanchor transitions. The Studio does allow for this right now but I'm honestly not sure what the synced code should be after such a thing (and the exported code seems to ignore such transition - so I'd say this is a bug). 

Putting aside the reanchoring thing - the `serializableId` is would quickly become confusing and a lie here when you start reparenting states.

farskid: `:serializableId:invocation[:index]` is the format the current extractor uses and Studio expects. As far as I can tell, no specific operation is done by the value of it so it can change.
Introducing parent id here is just temporary so that I had a way to ensure the block is extracted in the right place. It can be removed too.
I'm unsure what happens to the id after re-anchoring in Studio but I can check. For now, this can be any value.

new-packages/ts-project/src/state.ts
Andarist: > or what we're doing to the actor id?

This. What makes actor id special? I don't recall a similar pattern used anywhere so far. Like, we didn't do anything like this to `history: someVariable`.

As I mentioned, I don't think this is bad - I just wonder if there is any specific reason/line of thought that made you to do this specifically here.

new-packages/ts-project/src/state.ts
Andarist: In a very similar manner even using `index` here is dubious. 

Simplified Input:
```ts
src: [a, b] // `b` gets assigned id with `[1]` in it
```

Studio actions:
1. remove `a`
2. add new invoke

Would those ever be updated? If not then we end up with a confusing thing because `b` is at the first position (index 0) but yet it uses `[1]` in its inline id

new-packages/ts-project/src/state.ts
farskid: I checked Studio and we don't update this value in any special way. The only thing that Studio cares about is that if user didn't provide any `id` to the actor, the id that's extracted must be in a special format so Studio can render it as `anonymous`.

Andarist: so this could be simplified to smth like this, right?
```ts
`inline:${uniqueId()}`
```

farskid: It can be anything that doesn't include `:invocation`, apparently. Below is the Studio code responsible for this check:
```ts
export const isUserOverrideId = (
  id: string | undefined,
): id is SerialisableInvocationId => {
  if (!id) return false;
  return !id.includes(':invocation');
};
```
But we can be consistent and use the same format we have already.

new-packages/ts-project/__tests__/utils.ts
```diff
@@ -149,6 +150,12 @@ export function replaceUniqueIds(
       ...Object.keys(digraph.blocks).map(
         (id, i) => [id, `block-${i}`] as const,
       ),
+      ...Object.values(digraph.blocks)
+        .filter((block): block is ActorBlock => block.blockType === 'actor')
+        .filter((block) => block.properties.id.startsWith('inline:'))
+        .map(
+          (block, i) => [block.properties.id, `inline:actor-id-${i}`] as const,
+        ),
```
Andarist, line 158: nit: this could be merged with the existing iteration over blocks with the help of `.flatMap`. It's a test code so whatever ;p

new-packages/ts-project/src/state.ts
farskid: https://github.com/statelyai/xstate-tools/pull/442/files#r1457510255

PR #443, opened 2024-01-18
Extract `on` transitions

Andarist:


new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -0,0 +1,2235 @@
...
+  const tmpPath = await testdir({
+    'tsconfig.json': JSON.stringify({}),
+    'index.ts': ts`
+      import { createMachine } from "xstate";
+
+      createMachine({
+        states: {
+          foo: {
+            on: {
+              FOO: { target: "bar" },
```
farskid, line 350: This example doesn't match test title.

new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -0,0 +1,2235 @@
...
+test('should extract transition with multiple targets (array of strings)', async () => {
+  const tmpPath = await testdir({
+    'tsconfig.json': JSON.stringify({}),
+    'index.ts': ts`
+      import { createMachine } from "xstate";
+
+      createMachine({
+        on: {
+          MULTIPLE: {
+            target: [".a.a2", ".b.b2"],
```
farskid, line 1130: nit: This test can be divided into two tests. One testing simple target strings in array and another testing deep child `.a.b.c`.

new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -0,0 +1,2235 @@
...
+                    "eventType": "MULTIPLE",
+                    "type": "named",
+                  },
+                  "guard": undefined,
+                  "internal": true,
+                },
+                "source": "state-0",
+                "targets": [
+                  "state-3",
+                  "state-6",
```
farskid, line 1179: Why do we have both of these states as the target?

new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -0,0 +1,2235 @@
...
+            },
+            "root": "state-0",
+          },
+          [],
+        ],
+      ]
+    `);
+});
+
+test('should extract forbidden transition (null)', async () => {
```
farskid, line 1405: This test is duplicated.

new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -0,0 +1,2235 @@
...
+            },
+            "root": "state-0",
+          },
+          [],
+        ],
+      ]
+    `);
+});
+
+test('should extract transition description', async () => {
```
farskid, line 2141: nit: if we're going to have a single test for description, I'd rather have it be a multiline description. We've had problems with those before.

farskid:
@Andarist Are we not extracting always and after transitions in this PR?

Andarist:
> @Andarist Are we not extracting always and after transitions in this PR?

I'm working on `always` right now. I plan to just add those in followup PRs to make each PR smaller and easier for review etc.

new-packages/ts-project/__tests__/transitions.test.ts
Andarist: The test title is:
> should extract transition with multiple targets (array of strings)

Those are two parallel regions that are targeted here by a single transition

new-packages/ts-project/__tests__/transitions.test.ts
Andarist: added a multiline one now

new-packages/ts-project/__tests__/transitions.test.ts
Andarist: fixed

new-packages/ts-project/__tests__/transitions.test.ts
farskid: Is that how Studio behaves too?

farskid: Quickly tested and this is not an edge Studio can handle. Honestly, multi target is a new concept to me.

farskid (approved):

new-packages/ts-project/src/state.ts
```diff
@@ -51,12 +63,72 @@ export function createActorBlock({
...
+  };
+}
+
+function isForbiddenTarget(
+  ts: typeof import('typescript'),
+  element: Expression,
+): boolean {
+  return (
+    isUndefined(ts, element) ||
+    // null isn't technically allowe by the XState's types but it behaves in the same way at runtime
```
farskid, line 97:
```suggestion
    // null isn't technically allowed by the XState's types but it behaves in the same way at runtime
```

new-packages/ts-project/__tests__/transitions.test.ts
Andarist: That's something to be fixed on the Studio's side.

PR #444, opened 2024-01-19
Extract guards

farskid:


new-packages/ts-project/src/state.ts
```diff
@@ -268,6 +290,34 @@ export function extractState(
                     'description',
                   );
 
+                  const guardV4 = findProperty(ctx, ts, element, 'cond');
+                  const guardV5 = findProperty(ctx, ts, element, 'guard');
+
+                  if (guardV4 && guardV5) {
+                    ctx.errors.push({
+                      type: 'property_mixed',
```
farskid, line 298: We can find a better name for this error.

new-packages/ts-project/src/state.ts
```diff
@@ -268,6 +290,34 @@ export function extractState(
...
+                        ? guard.initializer.text
+                        : `inline:${uniqueId()}`,
+                      parentId: node.uniqueId,
+                    });
+                    ctx.digraph.blocks[guardBlock.uniqueId] = guardBlock;
+                    ctx.digraph.implementations.guards[guardBlock.sourceId] ??=
+                      {
+                        type: 'guard',
+                        id: guardBlock.sourceId,
+                        name: guardBlock.sourceId,
```
Andarist, line 317: could you remind me why we need both? 🤔 who they are used by the Studio

new-packages/ts-project/src/state.ts
```diff
@@ -268,6 +290,34 @@ export function extractState(
...
+                  if (guardV4 && guardV5) {
+                    ctx.errors.push({
+                      type: 'property_mixed',
+                    });
+                  }
+                  const guard = guardV5 ?? guardV4;
+
+                  let guardBlock;
+
+                  if (guard) {
```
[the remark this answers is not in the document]
farskid: Ah, I knew I forgot something.

new-packages/ts-project/src/state.ts
farskid: I'm planning to open a separate PR to tweak block names when inline.

new-packages/ts-project/src/state.ts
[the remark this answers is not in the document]
farskid: Pushed.

new-packages/ts-project/src/state.ts
```diff
@@ -268,6 +286,65 @@ export function extractState(
...
+                        ts.isStringLiteralLike(typeProperty.initializer)
+                      ) {
+                        guardBlock = createGuardBlock({
+                          sourceId: typeProperty.initializer.text,
+                          parentId: node.uniqueId,
+                        });
+                      } else {
+                        ctx.errors.push({
+                          type: 'transition_property_unhandled',
+                        });
```
farskid, line 328: This is where an invalid value for the `type` property will raise an error. We can change it either here or later.

new-packages/ts-project/src/state.ts
```diff
@@ -232,8 +266,71 @@ function extractEdgeGroup(
...
+                ctx.errors.push({
+                  type: 'property_mixed',
+                });
+              }
+              seenGuardProp = true;
+
+              if (ts.isStringLiteralLike(prop.initializer)) {
+                const block = createGuardBlock({
+                  sourceId: prop.initializer.text,
+                  parentId: sourceId,
```
Andarist, line 297: why are we setting `node.uniqueId` as `parentId` here? it looks slightly off - how this is important for the Studio? shouldn't a guard be linked to a concrete edge rather than a node?

farskid: It should be the edge but we didn't have edges at the time of opening this PR :) I need to update it now.

new-packages/ts-project/src/types.ts
```diff
@@ -76,6 +79,14 @@ export interface ActorBlock extends BlockBase {
   };
 }
 
+export interface GuardBlock extends BlockBase {
+  blockType: 'guard';
+  properties: {
+    type: string;
+    params: {};
```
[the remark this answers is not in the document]
farskid: Yes. I wanted to add them in a separate PR but I'll quickly add them now.

new-packages/ts-project/src/state.ts
farskid: Done

new-packages/ts-project/src/types.ts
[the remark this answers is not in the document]
farskid: You're right. I remembered we added action parameters but I wonder if I'm hallucinating or did we accidentally lose that work.

PR #446, opened 2024-01-19
Extract transition actions

Andarist:


new-packages/ts-project/src/state.ts
```diff
@@ -124,6 +122,64 @@ function getObjectTransitionTargets(
...
+            sourceId: typeProperty.initializer.text,
+            parentId,
+          });
+        }
+      }
+
+      return createActionBlock({
+        sourceId: `inline:${uniqueId()}`,
+        parentId,
+      });
```
farskid, line 162: This means an invalid value for type property in a parameterized action will be extracted as an inline action. This is fine as long as it is our intention.

Andarist: I don't think that this is the case. Parameterized actions are extracted in the conditional block above this and when `typeProperty` is not found it `return`s

farskid: I mean for a case like `action: {type: {}}`, type property will be found but it's not a string so the code continues and extract it as an inline action whereas this is invalid.

Andarist: Ah, I see what you mean - thanks for pointing this out. This should definitely be reported as an error.

farskid (approved):
Looks good. Let's report an error for invalid type property value and merge.

Andarist:
I decided to raise an error but continue to extract it as inline. If we simply skip it we'll have a lot of problems with unsynchronized positions in the actions array etc. Cases like this won't work perfectly anyway - we need to figure out the exact DX that we want to deliver here. Right now all of the errors that we raise act as TODOs - those are just places we need to revisit later when tightening up the design and DX.

PR #447, opened 2024-01-19
Extract `invoke.onDone` and `invoke.onError`

Andarist:


new-packages/ts-project/src/state.ts
```diff
@@ -339,329 +322,323 @@ export function extractState(
...
-          if (!ts.isPropertyAssignment(prop)) {
-            ctx.errors.push({ type: 'state_property_unhandled' });
+        for (const transition of prop.initializer.properties) {
+          if (!ts.isPropertyAssignment(transition)) {
+            ctx.errors.push({ type: 'transition_property_unhandled' });
+            return;
+          }
+          const event = getPropertyKey(ctx, ts, transition);
+          if (!event) {
+            ctx.errors.push({ type: 'transition_property_unhandled' });
```
farskid, line 380: nit: maybe we wan use `forEachStaticPropery` here too.

farskid:
@Andarist onDone and onError actions are not extracted. Are they being added in a follow up?

farskid (approved):
Works great!

Andarist:
> @Andarist onDone and onError actions are not extracted. Are they being added in a follow up?

They are extracted from what I can see - I have just pushed out an extra test case to be extra sure.

new-packages/ts-project/src/state.ts
Andarist: done

PR #448, opened 2024-01-22
Transition meta

farskid:


new-packages/ts-project/src/state.ts
```diff
@@ -255,6 +256,30 @@ function extractEdgeGroup(
                 : undefined;
               return;
             }
+            case 'meta': {
+              if (ts.isObjectLiteralExpression(prop.initializer)) {
```
Andarist, line 260: That has to be handled in the same exact way as `state.meta` so all of this should get moved to a shared utility. TBH though - I have no idea what `transition.meta` is about. This information is quite hard to get at runtime in XState ;p

davidkpiano: It made more sense when we had `state.transitions`, although I understand why that was removed (to make state objects equal if the state hasn't changed). Still, it's useful for visualizing and inspecting the machine.

farskid: A shared utility makes sense.

PR #450, opened 2024-01-22
Extract delayed transitions

farskid:


new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -3034,3 +3034,392 @@ test('should extract transition description (multi-line)', async () => {
...
+          "data": {
+            "context": {},
+          },
+          "edges": {
+            "edge-0": {
+              "data": {
+                "actions": [],
+                "description": undefined,
+                "eventTypeData": {
+                  "delay": "100",
```
Andarist, line 3073: I wonder - does Studio try to auto-detect if this is a number or a string?

new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -3034,3 +3034,392 @@ test('should extract transition description (multi-line)', async () => {
...
+    'tsconfig.json': JSON.stringify({}),
+    'index.ts': ts`
+      import { createMachine } from "xstate";
+
+      createMachine({
+        initial: "foo",
+        states: {
+          foo: {
+            after: {
+              "named delay": "bar",
```
Andarist, line 3160: I'd replace this with:
```suggestion
              myDelay: "bar",
```

and add an extra test case with a property like the one u have here

new-packages/ts-project/__tests__/transitions.test.ts
farskid: It does but for formatting and displaying purposes. The delay that's stored in digraph definition is always a string.

new-packages/ts-project/__tests__/transitions.test.ts
farskid: I don't understand. Are you suggesting adding a third test for an identifier delay?

new-packages/ts-project/__tests__/transitions.test.ts
```diff
@@ -3838,3 +3838,504 @@ test('should raise error when transition.meta contains any value other than a pl
...
+            },
+          },
+          "root": "state-0",
+        },
+        [],
+      ],
+    ]
+  `);
+});
+test('should extract after transition (string delay)', async () => {
```
Andarist, line 4065:
```suggestion
test('should extract after transition (string with whitespace delay)', async () => {
```

Andarist (approved):

new-packages/ts-project/__tests__/transitions.test.ts
Andarist: ye, basically a test for the simplest test case (identifier, no whitespace) and a test for the more specific situation (string with whitespace)

new-packages/ts-project/__tests__/utils.ts
```diff
@@ -98,7 +98,7 @@ async function createTestProgram(
     );
   }
 
-  const host = ts.createCompilerHost(options);
+  const host = ts.createCompilerHost(options, true);
```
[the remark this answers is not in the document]
farskid: Let's revert it whenever later. Easy to do at any point.

PR #453, opened 2024-01-23
Extract `reenter` (and old `internal`) transition property

Andarist:
tbh, it's hard to tell how it will **exactly** work with the Studio since it permits `undefined`. On top of that, it's not completely obvious how this property "evolves" when reanchoring transitions. For many transitions this property doesn't matter (only transitions that target descendants differ in behavior based on this property). I think we'll have to learn in practice how this plays out and tweak things if needed

### DIFF
--- a/new-packages/ts-project/src/index.ts
+++ b/new-packages/ts-project/src/index.ts
@@ -5,6 +5,7 @@ import {
   ExtractionError,
   ExtractorDigraphDef,
   TreeNode,
+  XStateVersion,
 } from './types';
 
 function findCreateMachineCalls(
@@ -104,31 +105,10 @@ function resolveTargets(ctx: ExtractionContext) {
 }
 
 function extractMachineConfig(
+  ctx: ExtractionContext,
   ts: typeof import('typescript'),
   createMachineCall: CallExpression,
-  sourceFile: SourceFile,
 ): readonly [ExtractorDigraphDef | undefined, ExtractionError[]] {
-  const ctx: ExtractionContext = {
-    sourceFile,
-    errors: [],
-    digraph: {
-      nodes: {},
-      edges: {},
-      blocks: {},
-      implementations: {
-        actions: {},
-        actors: {},
-        guards: {},
-      },
-      data: {
-        context: {},
-      },
-    },
-    treeNodes: {},
-    idMap: {},
-    originalTargets: {},
-  };
-
   const rootState = createMachineCall.arguments[0];
   const rootNode = extractState(ctx, ts, rootState, undefined);
 
@@ -154,6 +134,7 @@ function extractMachineConfig(
 export function createProject(
   ts: typeof import('typescript'),
   tsProgram: Program,
+  { version = 'v5' }: { version?: XStateVersion },
 ) {
   return {
     extractMachines(fileName: string) {
@@ -162,7 +143,28 @@ export function createProject(
         return [];
       }
       return findCreateMachineCalls(ts, sourceFile).map((call) => {
-        return extractMachineConfig(ts, call, sourceFile);
+        const ctx: ExtractionContext = {
+          sourceFile,
+          version,
+          errors: [],
+          digraph: {
+            nodes: {},
+            edges: {},
+            blocks: {},
+            implementations: {
+              actions: {},
+              actors: {},
+              guards: {},
+            },
+            data: {
+              context: {},
+            },
+          },
+          treeNodes: {},
+          idMap: {},
+          originalTargets: {},
+        };
+        return extractMachineConfig(ctx, ts, call);
       });
     },
   };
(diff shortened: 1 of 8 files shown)


new-packages/ts-project/src/index.ts
```diff
@@ -162,7 +147,28 @@ export function createProject(
         return [];
       }
       return findCreateMachineCalls(ts, sourceFile).map((call) => {
-        return extractMachineConfig(ts, call, sourceFile);
+        const ctx: ExtractionContext = {
+          sourceFile,
+          version,
```
farskid, line 152: Can we call this more specifically like `xstateVersion`?

new-packages/ts-project/src/state.ts
```diff
@@ -221,6 +222,16 @@ function registerActionBlocks(
   }
 }
 
+function getDefaultInternalValue(
+  ctx: ExtractionContext,
+  targets: string[] | undefined,
+) {
+  if (ctx.version !== 'v4') {
```
farskid, line 229: Why isn't the version a number? `version: 4 | 5`.

new-packages/ts-project/src/state.ts
```diff
@@ -221,6 +222,16 @@ function registerActionBlocks(
   }
 }
 
+function getDefaultInternalValue(
```
farskid, line 225: This function could use a code comment as to what we mean by `default internal value`.

new-packages/ts-project/src/state.ts
```diff
@@ -348,6 +361,43 @@ function extractEdgeGroup(
               registerActionBlocks(ctx, blocks, edge.data.actions);
               return;
             }
+            case 'internal': {
+              if (seenInternalProp) {
+                // `reenter` was already seen
```
farskid, line 366: Where was this `reenter` seen in this case?

new-packages/ts-project/src/state.ts
```diff
@@ -348,6 +361,43 @@ function extractEdgeGroup(
...
+                ctx.errors.push({
+                  type: 'property_mixed',
+                });
+                return;
+              }
+              seenInternalProp = true;
+
+              if (findProperty(ctx, ts, element, 'reenter')) {
+                return;
+              }
```
farskid, line 376: Shouldn't we raise a property_mixed error in this case too? I'm confused about the difference between this line of code and the one above that checks for `seenInternalProp`.

new-packages/ts-project/src/index.ts
Andarist: sure

new-packages/ts-project/src/state.ts
```diff
@@ -348,6 +361,43 @@ function extractEdgeGroup(
...
+
+              const value = getJsonValue(ctx, ts, prop.initializer);
```
farskid, line 378: nit: I'd rather use `getJsonValue` for when we want to extract a JSON value like we do with metadata, context, etc. We could directly check for AST kind here which is more specific IMO.

new-packages/ts-project/src/state.ts
Andarist: it just felt nicer to me to be able to potentially extend this in the future with stuff like `'v5.10'` instead of `5.1` (`v5.10` is not representable as a number since it's the same as `5.1`)

farskid: Are you thinking a runtime/extraction difference in non-major XState versions?

new-packages/ts-project/src/state.ts
Andarist: good point, gonna do this change

new-packages/ts-project/src/state.ts
Andarist: we'll raise it when we reach `reenter` - not raising it here avoids it being raised twice, IIRC. I agree that this is slightly weirdly written but remember that each of those can be seen first - so there is a little bit of a dance happening here to avoid reporting it twice and stuff.

new-packages/ts-project/src/state.ts
Andarist: that might happen with:
```ts
internal: true,
reenter: true
```

whereas the other "branch" gets hit with:
```ts
reenter: true,
internal: true
```

new-packages/ts-project/src/state.ts
Andarist: yes

Andarist: the runtime/extraction behavior etc should rather stay the same within the major but we might want to raise errors differently or something

farskid: I'd drop the v because it's given by the property name but this is nit.

farskid (approved):

PR #460, opened 2024-02-01
Collect AST paths of nodes and edges

Andarist:


new-packages/ts-project/__tests__/actors.test.ts
```diff
@@ -110,22 +110,22 @@ test('should extract multiple actors with different string sources (direct)', as
             "blocks": {
               "block-0": {
                 "blockType": "actor",
-                "parentId": "state-1",
+                "parentId": "state-2",
```
Andarist, line 113: Those changes are related to the insertion order of nodes. Previously we were iterating through states using a regular `for-of` loop but now I'm using `forEachStaticProperty` and that one iterates in reverse.

So blocks belonging to states appearing later within its parents are coming first now. I brought back the original order of states by sorting them based on the created `getNodeSourceOrders` utility. I mainly have done it to reduce the noise here and to make those snapshots slightly more readable. I think it's not such a big issue for other types (like blocks and edges). It would be cool to sort them in a similar way but I don't think it's worth the added complexity (and time) right now.

Alternatively, I could not use `forEachStaticProperty` and push the AST segments onto the `currentAstPath` manually~. Maybe I will pick this solution in a sec, I'm not sure yet.

As mentioned, the ordering problem was not unique to nodes - edges were and are "misordered" this way already. All of this is just purely a snapshot issue and not a functional one ;p

farskid (approved):

PR #462, opened 2024-02-02
Correct actor id

farskid:


new-packages/ts-project/src/state.ts
```diff
@@ -793,7 +793,7 @@ export function extractState(
                   ? srcProperty.initializer.text
                   : `inline:${uniqueId()}`,
                 parentId: node.uniqueId,
-                actorId: actorId ?? `inline:${uniqueId()}`,
+                actorId: actorId ?? `inline:invocation[${index}]`,
```
Andarist, line 796:
```suggestion
                actorId: actorId ?? `inline:${uniqueId()}:invocation`,
```

new-packages/ts-project/src/state.ts
```diff
@@ -822,7 +822,7 @@ export function extractState(
             return createActorBlock({
               sourceId: `inline:${uniqueId()}`,
               parentId: node.uniqueId,
-              actorId: `inline:${uniqueId()}}`,
+              actorId: `inline:invocation[${index}]`,
```
Andarist, line 825:
```suggestion
              actorId: `inline:${uniqueId()}:invocation`,
```

new-packages/ts-project/__tests__/utils.ts
```diff
@@ -160,12 +160,6 @@ export function replaceUniqueIds(
       ...Object.keys(digraph.blocks).map(
         (id, i) => [id, `block-${i}`] as const,
       ),
-      ...Object.values(digraph.blocks)
-        .filter((block): block is ActorBlock => block.blockType === 'actor')
-        .filter((block) => block.properties.id.startsWith('inline:'))
-        .map(
-          (block, i) => [block.properties.id, `inline:actor-id-${i}`] as const,
-        ),
```
Andarist, line 168: I'd prefer to bring back this code - we can't have all implied actor ids to be the same. All of them has to include some unique part (that's why I'm proposing to bring back the `uniqueId` in them)

farskid: Yeah, that's fine. Studio however needs to add some uniqueness to the generated actor IDs too because it only relies on index. I can add that later.

PR #466, opened 2024-02-07
Implement initial state edits

Andarist:


new-packages/ts-project/src/utils.ts
```diff
@@ -179,7 +179,7 @@ export function everyDefined<T>(arr: T[]): arr is NonNullable<T>[] {
 }
 
 export function findProperty(
-  ctx: ExtractionContext,
+  ctx: ExtractionContext | undefined,
```
Andarist, line 182: I think it makes sense to have this conditional. Sometimes we want to raise "soft" errors because we are just extracting the digraph and sometimes we don't want to raise errors at all OR we want to throw a hard error because some invariant has not been met (like when we want to edit the initial property but that doesn't exist).

This might change in the future, maybe we should move this to the last argument or maybe we should even just provide something like `reportError` instead of the whole `ctx`. Either way, I think it's more than fine to refactor this later - once we learn more about the needs of those functions etc.

new-packages/ts-project/__tests__/source-edits/set-initial-state.test.ts
```diff
@@ -0,0 +1,135 @@
...
+          },
+        },
+      },
+    });",
+    }
+  `,
+  );
+});
+
+test.todo('should add initial state to root', async () => {
```
Andarist, line 106: I'll prepare a separate PR for this since it introduces more complexity related to adding new properties to the object, detecting the desired indentation etc.

farskid (approved):

Andarist:
Landed as https://github.com/statelyai/xstate-tools/commit/596c9a73429d73105915ae1c52869929f086173d

PR #472, opened 2024-02-22
Handle state descriptions

Andarist:


new-packages/ts-project/__tests__/source-edits/set-description.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    });",
+    }
+  `);
+});
+
+test('should be possible to remove a description at the root', async () => {
+  const tmpPath = await testdir({
+    'tsconfig.json': JSON.stringify({}),
+    'index.ts': ts`
+      import { creteMachine } from "xstate";
```
davidkpiano, line 120:
```suggestion
      import { createMachine } from "xstate";
```

new-packages/ts-project/__tests__/source-edits/set-description.test.ts
```diff
@@ -0,0 +1,184 @@
...
+      machineIndex: 0,
+    },
+    {
+      type: 'set_description',
+      statePath: [],
+    },
+  );
+  expect(await project.applyTextEdits(textEdits)).toMatchInlineSnapshot(`
+    {
+      "index.ts": "import { creteMachine } from "xstate";
```
davidkpiano, line 142:
```suggestion
      "index.ts": "import { createMachine } from "xstate";
```

new-packages/ts-project/__tests__/source-edits/set-description.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    });",
+    }
+  `);
+});
+
+test('should be possible to add a multiline description', async () => {
+  const tmpPath = await testdir({
+    'tsconfig.json': JSON.stringify({}),
+    'index.ts': ts`
+      import { creteMachine } from "xstate";
```
davidkpiano, line 154:
```suggestion
      import { createMachine } from "xstate";
```

new-packages/ts-project/__tests__/source-edits/set-description.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    },
+  );
+  expect(await project.applyTextEdits(textEdits)).toMatchInlineSnapshot(`
+    {
+      "index.ts": "import { creteMachine } from "xstate";
+
+    createMachine({
+    	description: \`wow
+    	so much
+    	wow\`
```
davidkpiano, line 180: Shouldn't this be this?
```suggestion
so much
wow\`
```

Andarist: Yes, and no. I was thinking about this just today because you are right, literal-wise - it should be like this. It gets somewhat complicated when moving descriptions around in a file though (like when reparenting states etc). I concluded that - for simplicity - we should ignore leading whitespace in each line of the description (in the Studio, I mean).

We could just not use template literals here but then longer multiline descriptions would look quite awful in the code.

In the future, I'd love to treat descriptions as comments - that would eliminate the concern since code comments wouldn't suffer from this leading whitespace problem.

davidkpiano: Hmm, perhaps we should do the safe thing and just not use template literals then. Descriptions don't just apply to the studio

Andarist: Avoiding template literals won't work perfectly either because people might be using template literals in other parts of the code (which also means that my idea of ignoring leading whitespace isn't that great either). I'll think about what we can do about it - the worst-case scenario is that we'll have to have a smarter indenter than `text.replace(/^/gm, indentation)`. 

Given the above... ye, I have to change the expected output here (in the way you have suggested)

davidkpiano (approved):

new-packages/ts-project/__tests__/source-edits/set-description.test.ts
Andarist: I'll work on fixing this later on - probably when I'll have to deal with reparenting nodes

PR #474, opened 2024-02-27
Handle basic transitions insertions

Andarist:


new-packages/ts-project/__tests__/source-edits/add-transition.test.ts
```diff
@@ -0,0 +1,1830 @@
...
+    {
+      type: 'add_transition',
+      sourcePath: [],
+      targetPath: null,
+      transitionPath: ['on', 'FOO', 0],
+    },
+  );
+  expect(await project.applyTextEdits(textEdits)).toMatchInlineSnapshot(`
+    {
+      "index.ts": "import { creteMachine } from "xstate";
```
davidkpiano, line 30: `crete` -> `create`

new-packages/ts-project/__tests__/source-edits/add-transition.test.ts
```diff
@@ -0,0 +1,1830 @@
...
+
+    createMachine({
+    	on: {
+    		FOO: undefined
```
davidkpiano, line 34: Should this be `FOO: {}` ?

new-packages/ts-project/__tests__/source-edits/add-transition.test.ts
Andarist: aaaaargh, i keep copy-pasting the same typo 😂

new-packages/ts-project/__tests__/source-edits/add-transition.test.ts
Andarist: It could be - it's kinda a stylistic difference that has no runtime impact. I find the `undefined` variant more explicit than an empty object. Especially that a diff like this looks kinda better (to me):
```diff
-FOO: undefined
+FOO: {
+  target: undefined,
+  guard: 'checkStuff',
+}
```
than
```diff
-FOO: {}
+FOO: {
+  target: undefined,
+  guard: 'checkStuff',
+}
```

Like, none of those above is bad - but it feels good to me that when adding a guard we just have to wrap the short syntax for the target into, well, target property + add a guard. The alternative is to add 2 properties to an object (target+guard) with the target not having a clear correlation with the previous version of the code. The current version has also this benefit that it works in the very same way with `FOO: 'someTarget'` and `FOO: undefined` - it just wraps that non-object value as `target` and doesn't have to special-case `FOO: {}` anyhow

Andarist: We are not dealing with serialization here at all - and even the serialized value  has still the same semantic meaning. In fact, the serialized value looks exactly the same as what you propose to output here

davidkpiano (approved):

